PR #20, opened 2017-07-22
Don't extend FileIO's load and save

### LINKED ISSUE #19: segfault when passing an IOStream to JLD2.load
After
```
a = sprand(100, 100, 0.5)
@save "testjld2" a
```

I tried to pass an IOStream to `load`:
```
f = open("testjld2", "r")
# IOStream(<file testjld2>)

JLD2.load(f)
Segmentation fault: 11
```

~It'd be great if JLD2.load could take an IOStream.~ The docs imply that it should work.

timholy:
Fixes #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The code here should work on Julia 0.6. It has extensive unit tests, but it has 
 The `@save` and `@load` macros are the simplest way to interact with a JLD2 file. The `@save` macro writes one or more variables from the current scope to the JLD file. For example:
 
 ```julia
+using JLD2, FileIO
 hello = "world"
 foo = :bar
 @save "example.jld2" hello foo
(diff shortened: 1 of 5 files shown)


timholy:
Obviously the negative here is that users now have to remember to say `using FileIO`.

README.md
```diff
@@ -15,6 +15,7 @@ The code here should work on Julia 0.6. It has extensive unit tests, but it has
 The `@save` and `@load` macros are the simplest way to interact with a JLD2 file. The `@save` macro writes one or more variables from the current scope to the JLD file. For example:
 
 ```julia
+using JLD2, FileIO
```
simonster: I don't think `FileIO` is necessary here, since the macro form is still provided by JLD2.

simonster:
Thanks. I just copied this code from JLD; I didn't realize that these weren't supposed to be extending FileIO.

README.md
timholy: True. I was just using it as another reminder, but if you prefer I'll get rid of it.

PR #27, opened 2017-08-06
Fix #26

### LINKED ISSUE #26: Problem reading dictionary
My MWE (**Edit:** import `Colors.jl` also):
```julia
import JLD2, Colors

function generator(size, β, runs, name)

    parameters = Dict(  "size" => size,
                        "beta" => β,
                        "runs" => runs)

    JLD2.jldopen(name, "w") do file
        file["parameters"] = parameters
    end
end

size, β, runs = 10, 1.0, 10
generator(size, β, runs, "issue.jld2")

JLD2.jldopen("issue.jld2", "r+") do file
    @show file["parameters"]
end
```
which throws
```
LoadError: field "second" of type Pair{String,Any} must be defined in the current workspace, but was undefined in the file
```

If I set instead `size, β, runs = 10, 10, 10`, I get wrong values:
```
file["parameters"] = Dict("runs"=>0,"size"=>0,"beta"=>0)
```
Am I missing something? 

Version info:
```julia
julia> Pkg.status("JLD2")
 - JLD2                          0.0.2+             master
julia> Pkg.status("Colors")
 - Colors                        0.8.0

julia> versioninfo()
Julia Version 0.6.1-pre.0
Commit dcf39a1dda (2017-06-19 13:06 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-4510U CPU @ 2.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)
```

simonster:
Replaces `sizeof` with `odr_sizeof` when the argument is an ODR. This is probably a good idea generally, since otherwise something weird might happen if someone redefined `sizeof` on their type.

As best as I can tell, what happened with #26 is that, after loading FixedPointNumbers, [this line](https://github.com/simonster/JLD2.jl/blob/3f700ab5b9beb0cda3e974136b3e62aacd867c8f/src/data.jl#L1332) would call `sizeof(Vlen{String})` and get 24 (which is what would be returned by the method defined in Base) rather than 16 (which is what should be returned by [this definition](https://github.com/simonster/JLD2.jl/blob/3f700ab5b9beb0cda3e974136b3e62aacd867c8f/src/data.jl#L21) earlier in the file, and is what is returned if you call `sizeof(JLD2.Vlen{String})` from the REPL). This made JLD2 skip 8 bytes between writing a string and writing the next field in an object. Apparently, when FixedPointNumbers defined a new method of `sizeof`, this made the generated function forget about the methods defined in JLD2. The errant behavior is sensitive to many things: precompilation must be enabled, JLD2 must be loaded before FixedPointNumbers, and `sizeof(JLD2.Vlen{String})` must not be called manually before it is called from the generated function. The documentation for generated functions is carefully written so that it is ambiguous whether or not this is a bug in Julia.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -34,7 +34,7 @@ end
 define_packed(AttributeHeader)
 
 Base.sizeof(attr::WrittenAttribute) = 8 + symbol_length(attr.name) + 1 + sizeof(attr.datatype) + sizeof(attr.dataspace) +
-                                     numel(attr.dataspace) * sizeof(objodr(attr.data))
+                                     numel(attr.dataspace) * odr_sizeof(objodr(attr.data))
 
 function write_attribute(io::IO, f::JLDFile, attr::WrittenAttribute, wsession::JLDWriteSession)
     namelen = symbol_length(attr.name)
--- a/src/data.jl
+++ b/src/data.jl
@@ -3,6 +3,8 @@ const Initialized = Union{Type{Val{true}}, Type{Val{false}}}
 const Pointers = Union{Ptr{Void}, IndirectPointer}
 
 struct OnDiskRepresentation{Offsets,JLTypes,H5Types} end
+odr_sizeof(::Void) = 0
+@Base.pure odr_sizeof(x::DataType) = Int(x.size)
 
 struct UnknownType{T}
     name::T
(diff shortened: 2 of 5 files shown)


timholy:
I'm not quite sure why that `sizeof` method was defined (by me, https://github.com/JuliaMath/FixedPointNumbers.jl/pull/59); the FPN tests pass on 0.6 without it.

PR #30, opened 2017-08-09
use the merge commit as VERSION cutoff for struct_type expr head

tkelman:
ref https://github.com/simonster/JLD2.jl/pull/28/files#r132098827


tkelman:
ping @simonster in case you have notifications from watching filtered differently than from direct mentions

PR #41, opened 2017-10-12
add get() and get!() for JLDFile and Group

rdeits:
This PR implements `get()` and `get!()` for JLDFile and Group. Following the example from Base's `dict.jl`, it implements:

```
get(f::Union{JLDFile, Group}, name::AbstractString, default)
get!(f::Union{JLDFile, Group}, name::AbstractString, default)  # mutates f to add the given key
get(default_generator::Function, f::Union{JLDFile, Group}, name::AbstractString) # calls default_generator() to generate the default value if `name` not found
get!(default_generator::Function, f::Union{JLDFile, Group}, name::AbstractString)  # mutates f to add the given key created by calling `default_generator()` if `name` not found
```

This is somewhat suboptimal because it will perform the key lookup twice: once to check if the key is present and once to actually retrieve the element. Is that OK? This could be improved as a future optimization without changing the public API.


simonster:
Thanks! I think performing two lookups is fine for now, since a hash table lookup is usually faster than IO.

PR #45, opened 2017-10-24
Fix on 0.7

timholy:
Fixes many deprecation warnings in addition to breakages. Tests pass on 0.6 and 0.7 locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - osx
 julia:
   - 0.6
+  - nightly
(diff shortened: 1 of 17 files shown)


timholy:
Test failure reported here: https://github.com/JuliaLang/julia/issues/24308

Also worth noting that while the "timed" tests take up to ~20s on Unix platforms, they are slow on AppVeyor (esp. 32-bit nightly).

simonster:
Thanks!

Issue #55, opened 2018-01-15
File saved with JLD2 can't be read, from identical OS and JLD2 version


Datseris:
Hello, I am facing a serious problem, namely that a file saved with `JLD2` version `0.0.5` and `FileIO` cannot be read from the same OS with the same version of `JLD2`. 

It is really weird for me because I have a bunch of files (computed with different parameters of a model) and most of them work perfectly.

Notice that for *all* of my files being computed, STDERR was empty, and STDOUT was exactly as expected (reporting some timings of computations and whatever).

I know paste my error messages:

![image](https://user-images.githubusercontent.com/19669089/34953714-f8d129d8-fa1d-11e7-8b71-7373d3abb9dc.png)

which does not work. However, another file, produced by the identical code with one parameter different (positive instead of negative energy):

![image](https://user-images.githubusercontent.com/19669089/34953743-0cccfdd6-fa1e-11e7-8175-7d127cfc6776.png)

I have no reason to believe that there is something wrong with the data of the first case, as I have done the computation for one sub-case and it work.

I am more than willing to share the 2 files presented here if they can help (I really need some help since I cannot load some long-running cluster results).

Datseris:
I have noticed that **all** my data that have negative energy cannot be loaded.

Can there be an issue with having `-` in the file name?

ararslan:
> Can there be an issue with having `-` in the file name?

If you save it as something like test.jld2 and try to load that, does it work?

Datseris:
Unfortunately to save it with a different name, I would have to re-create the data and thus submit a lot of hours of cluster work.

Simply renaming the file did not fix the issue however. I will try to produce a similar dataset with a much smaller system on my laptop in the mean-time.

ararslan:
If renaming the file didn't work then I wouldn't bother generating a new file just to test this. I think we can rule out an issue with the file name.

Datseris:
I have re-run all my simulations, made extra care to be sure that JLD2 version is everywhere 0.0.5 and now I can load the data.

I wonder whether this was an issue on my end or something weird happened somehow. I am closing this assuming it was an issue on my end. If you think ottherwise you can reopen.

Datseris:
@ararslan Hello, with many tears I return to this issue once more.

It happened to me again, and once again I can't load my data... :(

Versions:
```
 - FileIO                        0.6.1
 - JLD2                          0.0.5
```

[...]

PR #62, opened 2018-02-26
add a more helpful error message for users that encounter #55

mweastwood:
This is my proposal for an expanded error message for users that encounter #55. This will point them in the right direction as opposed to the original terse `InvalidDataException()` message.

It would be nice to have a `jldopen(..., mmap=false)` method though :)

This is what it looks like in the wild:
```
julia> file["0142"] # try to read the missing object in my file
ERROR: InvalidDataException
Some network file systems do not properly implement mmap. As a result of
this, you may have experienced data loss. JLD2 may be directed to not use
mmap as follows:

    jldopen("file.jld2", true, true, true, IOStream) do file
        file["object"] = object
    end

Stacktrace:
 [1] mmap_may_have_failed() at /home/mweastwood/julia-packages/v0.6/JLD2/src/JLD2.jl:45
 [2] read_obj_start(::JLD2.MmapIO) at /home/mweastwood/julia-packages/v0.6/JLD2/src/object_headers.jl:47
 [3] isgroup(::JLD2.JLDFile{JLD2.MmapIO}, ::JLD2.RelOffset) at /home/mweastwood/julia-packages/v0.6/JLD2/src/object_headers.jl:75
 [4] getindex(::JLD2.Group{JLD2.JLDFile{JLD2.MmapIO}}, ::String) at /home/mweastwood/julia-packages/v0.6/JLD2/src/groups.jl:103
 [5] getindex(::JLD2.JLDFile{JLD2.MmapIO}, ::String) at /home/mweastwood/julia-packages/v0.6/JLD2/src/JLD2.jl:350
```

Thoughts/comments?

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -24,8 +24,26 @@ const FILE_HEADER = "$(REQUIRED_FILE_HEADER)$(CURRENT_VERSION)\x00 (Julia $(VERS
 
 struct UnsupportedVersionException <: Exception end
 struct UnsupportedFeatureException <: Exception end
-struct InvalidDataException <: Exception end
 struct InternalError <: Exception end
+struct InvalidDataException <: Exception msg::String end
+
+InvalidDataException() = InvalidDataException("")
+
+Base.show(io::IO, exception::InvalidDataException) =
+    print(io, "InvalidDataException\n", exception.msg)
+
+@noinline function mmap_may_have_failed()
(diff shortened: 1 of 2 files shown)


janEbert:
I like it, had this exact problem as well. As long as #55 is not fixed, this is a nice temporary warning that would have saved me some time. :)

Issue #61, opened 2018-02-20
JLD2 not installing under latest 0.7-DEV


StephenVavasis:
I'm trying to port my code from 0.6.2 to 0.7-DEV (0-day old master) and so far I haven't been able to install either JLD or JLD2 (Ubuntu 17.10).  The installation of JLD2 is failing because the code is apparently not able to find libz.  Below is the error message.

```
[ Info: Precompiling module JLD2
┌ Warning: Deprecated syntax `type` at /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:62.
│ Use `mutable struct` instead.
└ @ nothing lowlevel.jl:62
┌ Warning: `is_windows` is deprecated, use `Sys.iswindows` instead.
│   caller = top-level scope
└ @ Core :0
WARNING: importing deprecated binding Base.Void into Libz.
WARNING: Base.Void is deprecated, use Nothing instead.
  likely near /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:62
WARNING: Base.Void is deprecated, use Nothing instead.
  likely near /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:62
WARNING: Base.Void is deprecated, use Nothing instead.
  likely near /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:62
WARNING: Base.Void is deprecated, use Nothing instead.
  likely near /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:62
ERROR: LoadError: LoadError: error compiling version: could not load library "libz"
libz.so: cannot open shared object file: No such file or directory
Stacktrace:
 [1] top-level scope
 [2] include at ./boot.jl:295 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1060
 [4] include at ./sysimg.jl:27 [inlined]
 [5] include(::String) at /home/vavasis/.julia/v0.7/Libz/src/Libz.jl:3
 [6] top-level scope
 [7] include at ./boot.jl:295 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:1060
 [9] include(::Module, ::String) at ./sysimg.jl:27
 [10] top-level scope
 [11] eval at ./boot.jl:298 [inlined]
 [12] top-level scope at ./<missing>:3
in expression starting at /home/vavasis/.julia/v0.7/Libz/src/lowlevel.jl:110
[...]

PR #66, opened 2018-03-28
Fix for SimpleVector unexported on 0.7

ararslan:
Fixes the specific issue noted in https://github.com/simonster/JLD2.jl/issues/61#issuecomment-376883882, which is due to `SimpleVector` being unexported in 0.7. cc @kafisatz

### DIFF
--- a/src/data.jl
+++ b/src/data.jl
@@ -1024,9 +1024,9 @@ end
 
 ## SimpleVectors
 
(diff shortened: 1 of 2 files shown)


kafisatz:
Thank you for fixing this! I appreciate the support, noting that I still cannot build JLD2 because of this https://github.com/bicycle1885/CodecZlib.jl/issues/17

PR #68, opened 2018-04-17
Make save macro supports append mode

iblis17:
I ran into this issue as well: https://github.com/JuliaIO/JLD.jl/issues/90
So, I want this macro to be more handy.

```julia
julia> x, y, z = 1:3
1:3

julia> @save "./test.jld2" y

julia> @load "./test.jld2"
1-element Array{Symbol,1}:
 :y

julia> @save "./test.jld2" mode = "r+" x z

julia> @load "./test.jld2"
3-element Array{Symbol,1}:
 :y
 :x
 :z

julia> x, y, z
(1, 2, 3)
```

### DIFF
--- a/src/loadsave.jl
+++ b/src/loadsave.jl
@@ -7,6 +7,22 @@ function jldopen(f::Function, args...; kws...)
     end
 end
 
+function _extractmode(xs)
+  mode = "w"
+  vars = sizehint!(Vector{Symbol}(), length(xs))
+  for x ∈ xs
+      if x isa Expr && x.head == :(=) && length(x.args) == 2 && x.args[] == :mode
+          mode = x.args[2]
+          @assert(mode ∈ ("r+", "w", "w+", "a", "a+"), "unsupport mode: $mode")
+      elseif x isa Symbol
+          push!(vars)
+      else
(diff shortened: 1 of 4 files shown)


test/loadsave.jl
```diff
@@ -84,3 +84,30 @@ d = Dict("params/p1" => 1,
...
+      @save f mode = "r+" y z
+    end
+    let
+      @load f x y z
+      @test (x, y, z) == (1, 2, 3)
+    end
+
+    # if unsupport mode
+    let x = 42
+      @test_throws ArgumentError @save f mode = "❓" x
```
iblis17: well, how can I test the exception from compile time?

src/loadsave.jl
```diff
@@ -7,13 +7,37 @@ function jldopen(f::Function, args...; kws...)
     end
 end
 
+function _extractmode(xs)
+  mode = "w"
```
ararslan: Should be 4 space indents (Julia standard) for consistency with the rest of this package.

JonasIsensee:
A more general approach to passing arguments is implemented in #198 .
What do you think?

JonasIsensee:
closing in favor of #198

PR #92, opened 2018-08-09
Improved 1.0 fixes

staticfloat:
With these changes, `JLD2.jl` can pass all tests on 1.0, with one exception: there appears to be an optimizer bug.  On my machine, the `rw.jl` tests fail on the reading portion of the final iteration of the loop (e.g. using `IOStream` as the io type, and using a compressed stream).  I get some weird `EOFError()` that is due to a header getting super random values.  I'm pretty sure this is an optimizer error, because as I was debugging this, I discovered that if I just add some side-effect-free calls, things work.  I tried looking at the difference between `@code_typed` and `@code_typed optimized=false`, but the difference is so huge I can't tell what the optimizer might be doing wrong.

Pinging @keno because you're probably interested in this.  The last commit in this stack is the single-line change that needs to be added to get this to work on my machine.


simonster:
Thanks for looking into this!

ararslan:
_Shakes fist at Windows_

staticfloat:
Alright that sorted windows locally, so hopefully we can merge and tag once this goes green!

staticfloat:
THE PARTY VAN IS LEAVING EVERYBODY GET ABOARD

ararslan:
image

ararslan:
yikes

PR #104, opened 2018-08-23
Delete project and manifest files

andreasnoack:
This will make CIBot work while we prepare to use the new package manager.


andreasnoack:
Would also be great with a bugfix release once this is in.

PR #143, opened 2019-07-25
[RFC] Global compress option

### LINKED ISSUE #133: Global `compress` option
At the moment compression is disabled by default, and since it's not possible to pass options to `@save` macro, all the JLD2 files created with `@save` are uncompressed (unnecessarily huge).

It would be nice to add some mechanism to set the global `compress` option that would be used as the default choice for `compress=` arg in `jldopen()`. E.g. `JLD2.set_compress(true/false)` (+ `JLD2.compress()` getter).

The alternative would be to support `compress=` option in `@save` macro, but that would interfere with the potential `var=expr` feature (#132).

alyst:
The PR adds `JLD2.COMPRESS` var that is used as the default value of `compress` keyarg in `jldopen()` call (`jldopen(..., compress=COMPRESS[])`).
As before, the compression is disabled by default, but the user can change it (`JLD2.COMPRESS[] = true`).
This is especially useful for `@save`, because this macro doesn't support specifying `jldopen()` options, so before this PR `@save` was able to generate uncompressed files only.

I'm not sure about the public interface to control `JLD2.COMPRESS`, though (hence RFC).
Directly assigning it looks straightforward, but dedicated public methods (e.g. `jldcompress(::Bool)`) is an alternative option.

The PR also adds `JLD2.COMPACT_STORAGE_THRESHOLD` (defaults to 8192).
In my case it helps to reduce the size of JLD2 file storing short (fewer rows) and wide (many columns) data frames.

Closes #133

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -205,6 +205,15 @@ Defaults to `false`.
 """
 const COMPRESS = Ref(false)
 
+"""
+Threshold for using the compact storage layout.
+Data items of smaller size will use the compact layout.
+Bigger data items will use either chunked or continuous layout depending on their data type.
+
+Defaults to 8192.
+"""
+const COMPACT_STORAGE_THRESHOLD = Ref(8192)
+
 const OPEN_FILES = Dict{String,WeakRef}()
 function jldopen(fname::AbstractString, wr::Bool, create::Bool, truncate::Bool, iotype::T=MmapIO;
                  compress::Bool=COMPRESS[], mmaparrays::Bool=false) where T<:Union{Type{IOStream},Type{MmapIO}}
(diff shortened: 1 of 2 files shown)


timholy:
I don't see anything wrong here, but I'm not useful when it comes to compression options.

DilumAluthge:
It would be good to get a review from someone familiar with the compression code and compression options.

DilumAluthge:
@alyst Can you rebase this on master, please?

JonasIsensee:
Hi @alyst ,

would #198 be an alternative for you?
Do you see any downsides to just improving the `@save` API?
That way we wouldn't have to rely on a global state.

JonasIsensee:
closing in favor of #198

PR #146, opened 2019-09-12
add regression test for JuliaIO#134

thisrod:
I started work on a bug report, then realised the bug had been fixed since the last release.  However, it strikes me as the kind of bug that's likely to regress.


timholy:
Thanks, @thisrod!

PR #170, opened 2019-12-21
Avoid using the broken unsafe_isdefined function

vtjnash:
Use the standard API—that's what it's for.


simonster:
This entire package should probably be deprecated. I believe this function was correct in Julia 0.6, but evidently no one ever updated things for Julia 0.7+.

PR #173, opened 2020-01-02
use full type path in type resolution

JeffBezanson:
Currently a non-deterministic bug is possible if two different modules have a type with the same name. `_resolve_type` drops the first path component (the top-level module) and instead searches all modules in hash order. This fixes it by just using the whole path. I believe this was introduced by 9ef1e0979b2b7b92e46a33d025b80bc4232bb611. If this looks good I'll add a test.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JLD2"
 uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
-version = "0.1.10"
+version = "0.1.11"
 
 [deps]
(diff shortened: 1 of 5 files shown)


JeffBezanson:
Hmm, this should probably also check that everything but the last component names a module.

DilumAluthge:
This looks good to me. I think the CI failure is unrelated - I restarted the job.

It would be good to have a regression test.

DilumAluthge:
I've gone ahead and bumped the version number so that we can tag a new release as soon as we merge this PR.

JeffBezanson:
Ok should be set now.

Issue #136, opened 2019-04-02
Segfault when trying to load folder


felixcremer:
When I accidentally try to use a path to an existing folder to load a variable julia crashes with a segmentation fault:

```julia
julia> @load "test/" bar

signal (11): Segmentation fault
in expression starting at no file:0
_write at ./pointer.jl:155 [inlined]
write at /home/crem_fe/.julia/packages/JLD2/KjBIK/src/misc.jl:27 [inlined]
save_group at /home/crem_fe/.julia/packages/JLD2/KjBIK/src/groups.jl:348
unknown function (ip: 0x7eff19711ae3)
jl_fptr_trampoline at /buildworker/worker/package_linux64/build/src/gf.c:1864
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2219
close at /home/crem_fe/.julia/packages/JLD2/KjBIK/src/JLD2.jl:367
#jldopen#31 at /home/crem_fe/.julia/packages/JLD2/KjBIK/src/loadsave.jl:6
jl_fptr_trampoline at /buildworker/worker/package_linux64/build/src/gf.c:1864
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2219
jldopen at /home/crem_fe/.julia/packages/JLD2/KjBIK/src/loadsave.jl:2
jl_fptr_trampoline at /buildworker/worker/package_linux64/build/src/gf.c:1864
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2219
do_call at /buildworker/worker/package_linux64/build/src/interpreter.c:323
eval_value at /buildworker/worker/package_linux64/build/src/interpreter.c:411
eval_stmt_value at /buildworker/worker/package_linux64/build/src/interpreter.c:362 [inlined]
eval_body at /buildworker/worker/package_linux64/build/src/interpreter.c:773
jl_interpret_toplevel_thunk_callback at /buildworker/worker/package_linux64/build/src/interpreter.c:885
unknown function (ip: 0xfffffffffffffffe)
unknown function (ip: 0x7eff0540a94f)
unknown function (ip: 0x6)
jl_interpret_toplevel_thunk at /buildworker/worker/package_linux64/build/src/interpreter.c:894
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:764
jl_toplevel_eval_in at /buildworker/worker/package_linux64/build/src/toplevel.c:793
[...]

PR #184, opened 2020-04-02
Fix segmentation fault when reading a file that doesn't exist

### LINKED ISSUE #183: Segmentation fault
The following code snippet produces a segmentation fault:
```
using JLD2

jfn, _ = mktemp()
jldopen(string(jfn, ".jld2"), "a+") do ds
    ds["1"] = 1.0
end

#jldopen(string(jfn, ".jld2"), "r") do jds
jldopen(jfn, "r") do jds
    println(jds["1"])
end
```
The fault looks like:
```

signal (11): Segmentation fault
in expression starting at /central/home/kpamnany/tmp/jld2bugmwe.jl:9
macro expansion at ./pointer.jl:65 [inlined]
read! at ./io.jl:681 [inlined]
#jldopen#9 at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/JLD2.jl:241
jldopen at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/JLD2.jl:203 [inlined]
jldopen at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/JLD2.jl:203 [inlined]
#jldopen#10 at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/JLD2.jl:293
jldopen at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/JLD2.jl:288
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
#jldopen#31 at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/loadsave.jl:2
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2141 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
jldopen at /central/home/kpamnany/CLIMA/.julia_depot/packages/JLD2/hB4ya/src/loadsave.jl:2
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1631 [inlined]
do_call at /buildworker/worker/package_linux64/build/src/interpreter.c:328
eval_value at /buildworker/worker/package_linux64/build/src/interpreter.c:417
eval_stmt_value at /buildworker/worker/package_linux64/build/src/interpreter.c:368 [inlined]
eval_body at /buildworker/worker/package_linux64/build/src/interpreter.c:778
jl_interpret_toplevel_thunk_callback at /buildworker/worker/package_linux64/build/src/interpreter.c:888
unknown function (ip: 0xfffffffffffffffe)
unknown function (ip: 0x2aaad839740f)
unknown function (ip: 0x6)
jl_interpret_toplevel_thunk at /buildworker/worker/package_linux64/build/src/interpreter.c:897
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:814
jl_parse_eval_all at /buildworker/worker/package_linux64/build/src/ast.c:873
jl_load at /buildworker/worker/package_linux64/build/src/toplevel.c:878
include at ./boot.jl:328 [inlined]
include_relative at ./loading.jl:1105
include at ./Base.jl:31
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
exec_options at ./client.jl:287
_start at ./client.jl:460
jfptr__start_2084.clone_1 at /central/software/julia/1.3.0/lib/julia/sys.so (unknown line)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
unknown function (ip: 0x401931)
unknown function (ip: 0x401533)
__libc_start_main at /lib64/libc.so.6 (unknown line)
unknown function (ip: 0x4015d4)
Allocations: 8768916 (Pool: 8766924; Big: 1992); GC: 9
Segmentation fault
```

kpamnany:
Fixes #183.

### DIFF
--- a/src/mmapio.jl
+++ b/src/mmapio.jl
@@ -34,6 +34,7 @@ if Sys.isunix()
         newptr = ccall(:jl_mmap, Ptr{Cvoid}, (Ptr{Cvoid}, Csize_t, Cint, Cint, Cint, Int64),
                        C_NULL, n, Mmap.PROT_READ | (io.write*Mmap.PROT_WRITE),
                        Mmap.MAP_SHARED, fd(io.f), 0)
+        systemerror("mmap", newptr == Ptr{Nothing}(-1))
         io.n = n
         io.curptr += newptr - oldptr
         io.startptr = newptr
--- a/test/loadsave.jl
+++ b/test/loadsave.jl
@@ -134,3 +134,7 @@ vect_read = jldopen(fn,"r") do f
 end
 @test all(findall(ismissing,vect) == findall(ismissing,vect_read))
 @test all( skipmissing(vect) .=== skipmissing(vect_read))
+
+# Issue #183
+jfn, _ = mktemp()
+@test_throws SystemError jldopen(jfn, "r")


timholy:
Thanks! Can you add a test that would have triggered the segfault?

timholy:
Still segfaulting. Let us know when you're happy with this.

kpamnany:
Ah. Turns out that `MmapIO` actually does throw a `SystemError` if the file doesn't exist. I wondered how come this simple check was missing.

The problem is that `mktemp()` creates the file so it does exist, but is empty (?). And now, the `read!` [here](https://github.com/JuliaIO/JLD2.jl/blob/master/src/JLD2.jl#L241) is seg-faulting and I don't really understand why.

JonasIsensee:
My commit seems to fix the problem by introducing a new check that tests whether the filesize is zero.

However I do not understand why the test failed. Locally it worked for the new test some old test seems to break due to these changes.

kpamnany:
@timholy: I think this is the right fix.

JonasIsensee:
also fixes #136

JonasIsensee:
Does anyone else have an opinion on this?
I'd say we can merge this.

I'm not a big fan of the error message and feel like we could add proper filepath assertions but that will be a PR in its own right.

kpamnany:
Fixed the conflict.

timholy:
Seems OK to me. Indeed the error message does not rate high on verbosity but perhaps it's not easy to clearly convey the precise nature of the problem at this stage? It seems useful to have this fix as the "last interceptor" and then perhaps we could add other checks at an earlier stage in the pipeline. We could modify the test to still check this particular fix.

kpamnany:
Haha, well, it's an `mmap` system error! We'd need to check `errno` for more I think.

JonasIsensee:
> Seems OK to me. Indeed the error message does not rate high on verbosity but perhaps it's not easy to clearly convey the precise nature of the problem at this stage? It seems useful to have this fix as the "last interceptor" and then perhaps we could add other checks at an earlier stage in the pipeline. We could modify the test to still check this particular fix.

Exactly, that is what I was thinking of.
I don't think it will be necessary to dissect the `errno`. (Would that even work across different OS?)
Currently you will get this error any time mmap fails even if mmap is not really the problem.
It should be relatively easy to check whether the give filepath 
 1) contains a filename (don't try to open a folder e.g. #136
 2) all directories in the path exist

When all those cases are taken care of, one could even wrap the mmap-open into a `try` block and use `IOStream` as a fallback.
(Which would of course be pointless if mmap is not actually the problem)

PR #196, opened 2020-07-04
Allow saving of 'recursive' immutable structs

### LINKED ISSUE #189: Can't save a Measurements.Measurement
JLD2 currently seems unable to handle values of type `Measurements.Measurement`:

```julia
using Measurements, JLD2
x = measurement(5.0, 2.0)
@save "out.jld2" x
```

result in

```
ERROR: StackOverflowError:
Stacktrace:
 [1] hasfielddata(::Any) at /user/.julia/packages/JLD2/2W9VX/src/data.jl:55
 [2] hasdata(::DataType) at /user/.julia/packages/JLD2/2W9VX/src/data.jl:63
 [3] hasfielddata(::Any) at /user/.julia/packages/JLD2/2W9VX/src/data.jl:56
 ... (the last 2 lines are repeated 18523 more times)
```

CC @giordano - Not sure if something has to be changed in JLD2 or Measurements.jl to fix this.

JonasIsensee:
When serializing immutable structs, `JLD2` attempts to generate an *in-line* representation for the given struct.
It does so by recursively trying to find the memory layout of all immutable fields.
This is neat for many applications but it fails when encountering recursive structs e.g.
```
struct RecursiveStruct
    x::RecursiveStruct
    RecursiveStruct() = new()
    RecursiveStruct(x) = new(x)
end
```

This PR modifies the relevant function `hasfielddata` and `hasdata` to look for such recursive structs.
The functions now track which types they already have encountered and terminate the inlineing 
when a type signature is encountered again.

closes #189

### DIFF
--- a/src/data.jl
+++ b/src/data.jl
@@ -48,19 +48,21 @@ writeas(T::Type) = T
 odr_sizeof(::ReadRepresentation{T,S}) where {T,S} = odr_sizeof(S)
 
 # Determines whether a specific field type should be saved in the file
-@noinline function hasfielddata(@nospecialize T)
+@noinline function hasfielddata(@nospecialize(T), encounteredtypes=DataType[])
     T === Union{} && return false
     !isconcretetype(T) && return true
     T = T::DataType
+    T in encounteredtypes && return true
+    push!(encounteredtypes, T)
     (T.mutable || T <: Type) && return true
-    hasdata(T)
+    hasdata(T, encounteredtypes)
 end
 
 # Determines whether a specific type has fields that should be saved in the file
-function hasdata(T::DataType)
+function hasdata(T::DataType, encounteredtypes=DataType[])
     isempty(T.types) && T.size != 0 && return true
     for ty in T.types
-        hasfielddata(writeas(ty)) && return true
+        hasfielddata(writeas(ty), encounteredtypes) && return true
     end
     false
 end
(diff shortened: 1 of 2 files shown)


giordano:
I can't comment on the implementation as I'm not familiar with the code of the package, but I tried this PR with `Measurements.jl` and it seems to work great:
```julia
julia> using JLD2, Measurements

julia> x = 3 ± 1
3.0 ± 1.0

julia> y = 2x
6.0 ± 2.0

julia> y - x - x
0.0 ± 0.0

julia> z = 6 ± 2
6.0 ± 2.0

julia> z - x - x
0.0 ± 2.8

julia> @save "output.jld2" x y z

julia> JLD2.jldopen("output.jld2") do io
           io["y"] - io["x"] - io["x"]
       end
0.0 ± 0.0

julia> JLD2.jldopen("output.jld2") do io
           io["z"] - io["x"] - io["x"]
       end
0.0 ± 2.8
```
{Un,}correlated variables are handled corrected when deserialised, too!

oschulz:
Thanks for this!

oschulz:
Would it be possible to have a new release of JLD2 that includes this? This functionality would be very helpful for us.

JonasIsensee:
Hi,
I'm trying to prepare a somewhat bigger release. 
In particular I would like to include #206 but I am not super happy with that PR just yet.

I'd say the release is no more than 2 weeks away. If I can't fix #206 until then, I'll put out a release without it.
Does that sound alright for you?

oschulz:
Thanks, that sounds great! Two weeks is not a problem.

PR #197, opened 2020-07-04
make benchmarks runnable

JonasIsensee:
I noticed that the benchmarks have not been updated for compatibility with julia > v1.0.

While in the long run it might be good to implement something using `PkgBenchmark`
this PR for now makes the benchmarks runnable again.


benchmark/benchmark.jl
```diff
@@ -33,19 +33,19 @@ function serialize_benchmark_read()
...
     show(stdout, MIME("text/plain"), @benchmark jld_benchmark_read())
     println("\n")
-    gc()
+    GC.gc()
     println(title, " Base.serialize")
     show(stdout, MIME("text/plain"), @benchmark serialize_benchmark_write($data))
     println("\n")
-    gc()
+    GC.gc()
     println(title, " Base.deserialize")
```
jaakkor2, line 49: More appropriate strings could be " JLD2 write",  " JLD2 read", " Serialization.serialize", " Serialization.deserialize"

JonasIsensee: fair enough, fixed

Issue #162, opened 2019-11-20
Roadmap to 1.0


DilumAluthge:
What needs to be done before we can release JLD2 version 1.0.0?

alyst:
It would be nice to address #133 or have some way to control the compression from `@save`.

It would also be nice to do time/filesize profiling in typical workloads, because I have an impression that JLD2 is not very fast, and the filesizes are larger than alternatives (serialization, *.RData* files).
Also, enabling compression didn't reduce the file sizes as much as I have hoped.

timholy:
Another thought would be implement some variant of [addrequire](https://github.com/JuliaIO/JLD.jl#types-and-their-definitions). I don't want to portray this as essential because no matter what, I think we have to support files written in older formats whenever we move to 1.0.

DilumAluthge:
All of those are nice and good idea, but none of them are breaking, right?

The only reason to delay the release of 1.0 is if there are breaking changes we want to make. Once we release 1.0 we can still make as many non-breaking changes as we want.

timholy:
I don't think any are breaking.

alyst:
Compression configuration is theoretically API-breaking. It's just nice to decide what to do with it before releasing 1.0. If the planned change would not break the API, it could be done after 1.0.
There could be some storage format tweaks that improve the performance (also data corruption issues like #163 may require format changes), it would be nice to introduce the changes before 1.0 as well.

alyst:
Also, #126 & #151 were non-breaking, but many JLD2 files created before 0.1.6 are unusable because of the wrong type names.
This is something that should be avoided for files created with JLD2 1.0+.

cossio:
Is there anything holding back a 1.0 release at this point?

JonasIsensee:
As an update on this:
The compression part of the library needs a reworking. This can't be done without breaking API changes.
That is the main thing holding back 1.0 at this point.

PR #198, opened 2020-07-05
implement better save macro

### LINKED ISSUE #132: var=expr support in `@save` macro
At the moment `@save` macro just assumes that whatever expression is passed is a name of session variable. It would be nice to teach `@save` to recognize `var=expr` expressions, so that it saves the results of `expr` calculation under `var` name.

### LINKED ISSUE #133: Global `compress` option

JonasIsensee:
Hi,

this PR implements a new `@save` macro.
As far as I can see it does not break any old behaviour but it allows for more flexibility.

New features are:
*Option passing*: 
```
hello = "world"
@save "test.jld2" {compress=true} hello
@save "test.jld2" {compress=true, mmaparrays=false} hello
```
- *Assignment syntax*
```
@save "test.jld2" bye=hello
@save "test.jld2" hello randomnumber=rand(10)
```
This is relevant for:
#162
closes #132
closes #133

What do you think?
Is this a good compromise in terms of syntax?


JonasIsensee:
This PR now has an additional commit that adds `iotype` as a kwarg to the interface.
In this way one can now write e.g.

```
@save "test.jld2" {iotype=IOStream} key=val

jldopen("test.jld2", "w"; iotype=IOStream) do f
    write(f, key, val)
end
```

biona001:
Hello, so what is the new syntax for saving something with `compress=true`?

JonasIsensee:
Hi
the new syntax is
```
hello = "world"
@save "test.jld2" {compress=true} hello
```
but so far this is only on master.
In about 2 weeks there should be a new release so that you can update JLD2 locally for the new features.

PR #199, opened 2020-07-15
fix #195, struct_type failure on 1.5

### LINKED ISSUE #195: Fails on 1.5
PkgEval indicates this package fails on the upcoming 1.5:

https://github.com/JuliaCI/NanosoldierReports/blob/680e2355355927aeac1c26154cc0aca38b5c13e6/pkgeval/by_hash/bf34e0a_vs_b5f13c3/logs/JLD2/1.5.0-beta1-c594048585.log

It's not clear exactly why.

JeffBezanson:
Should be compatible with all versions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JLD2"
 uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
-version = "0.1.13"
+version = "0.1.14"
 
 [deps]
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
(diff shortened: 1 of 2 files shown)


JonasIsensee:
Ah, this is great!
Just so I understand this correctly:
Your `struct` expression can only eval to give a struct, right?

I'm wondering whether a malicious file could have a modified fieldtype that ends the struct definition and inserts arbitrary code.

(this is likely just hypothetical since the field types are probably checked beforehand but I'd have to look that up)

JeffBezanson:
Yes, if the names are Symbols and the types are all Types it should be fine.

PR #212, opened 2020-08-11
[WIP] Inline union repr for vectors

### LINKED ISSUE #35: Adapt JLD2 for isbits union fields
As implemented in JuliaLang/julia#22441. At the moment, this is impossible, because Julia does not correctly construct these objects. But once it does, there is the question of how these should be written in the file. Obvious options are:

- Write union-typed fields the same way we do now, by saving the field content in its own dataset. This might be easier for non-JLD2 implementations to consume, since it's a closer match with the HDF5 data model, but writing a dataset has substantially overhead in terms of both space and time vs. how Julia handles this in memory. It would be nice if the cost of saving/loading data with JLD2 was closely related to the cost of working with said data in Julia.
- Create datatypes for isbits unions. HDF5 doesn't actually support unions, but we could save a datatype structured like:

```
DATATYPE "00000002" H5T_COMPOUND {
         H5T_STD_I64LE "Int64";
         H5T_IEEE_F64LE "Float64";
         H5T_STD_U8LE "uniontype";
      }
```

where only one of the fields will be initialized, and the `ty` field will say which one. In principle this wastes some space in the file since only the `Int64` or `Float64` field will contain data, but the storage overhead is smaller than the storage overhead from creating a new dataset, and in the likely common case of `Union{T,Null}`, we would just be storing an extra byte to signal whether the value was null or not.

This would be a breaking change in that older versions of JLD2 wouldn't be able to read files created with newer versions, although newer versions of JLD2 should still be able to handle files created with older versions.

JonasIsensee:
This PR implements special cased storage for `Vector{T}` where `T` is a `Union` with two isbits member types.

In particular this allows efficient storage of `Vector{Union{Float64, Missing}}` and similar
by converting to a custom type `InlineUnionArray` before storing.

The overloads necessary are not all very pretty but that is needed since
the regular machinery only allows for elementwise special casing in Vectors.

closes #35 

At the moment the conversion is done element-wise.
However, AFAICT, julia already stores these arrays in a similar way in memory so possibly there could be a faster conversion.

Needs tests.
(Also, does this work when the vectors are fields in structs?)

closes #35


JonasIsensee:
It'd be really cool if this inline representation were transparent so that with any other HDF5 library you could still load the 2 data arrays and the mask

JonasIsensee:
So this even works when the union arrays are part of other structs.
Here a `DataFrame`
```
using DataFrames, JLD2
df = DataFrame(:a => rand(10^6))
@save "dfconcrete.jld2" df
df = DataFrame(:a => Union{Missing, Float64}[rand(10^6);])
@save "dfunioninline.jld2" df

# and using JLD2 0.1.14
df = DataFrame(:a => Union{Missing, Float64}[rand(10^6);])
@save "dfunionref.jld2" df
```
The resulting file sizes speak for themselves..
```
7,7M 13. Aug 09:51  dfconcrete.jld2
8,6M 13. Aug 09:53  dfunioninline.jld2
70M 13. Aug 09:54  dfunionref.jld2
```

With less interesting data but with compression:
```
jldopen("dfunioncompressed.jld2", "w"; compress=true) do f
       f["df"] = DataFrame(:a => Union{Missing, Int}[zeros(10^6);])
end
jldopen("dfunionnotcompressed.jld2", "w") do f
       f["df"] = DataFrame(:a => Union{Missing, Int}[zeros(10^6);])
end
```
```
17K 13. Aug 16:22  dfunioncompressed.jld2
8,6M 13. Aug 16:23  dfunionnotcompressed.jld2
```

JonasIsensee:
Really, the question here is:
Should we store these arrays the I implemented it now?
Alternatively it should be possible to store the mask+values in an interleaved fashion
so that we can store this as a HDF5 array with a custom compound type encoding the union.

The second option seems to be harder to implement in terms of the dispatch rules but should also be possible.

JonasIsensee:
See branch `inlineunion2` for an implementation of interleaved unions

JonasIsensee:
closing in favor of #221

PR #213, opened 2020-08-12
Magic Bytes

### LINKED ISSUE #211: Magic bytes
Hi,
the changes that I want to make in #206 modify the format in that older JLD2 versions cannot read some newer files.
(Old files can still be read)
Beside bumping the version number of the package, I feel we should also increase the version number 
of the file format in the magic byte file header. (e.g. move from v0.2 to v0.3)

However I just noticed that `JLD` and `JLD2` share essentially the same magic bytes:
```
# data formats
add_format(format"JLD", (unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), version 0.0"),
                         unsafe_wrap(Vector{UInt8}, "Julia data file (HDF5), version 0.1")), ".jld", [:JLD])
add_format(format"JLD2", "Julia data file (HDF5), version 0.2", ".jld2", [:JLD2])
```
and only differ by a version number.

This probably made sense a few years ago when JLD2 was still meant as a replacement for JLD but that is not very likely as of today.
What do you think? (Will JLD ever need higher version numbers?)
Should I still just add v0.3 for the magic bytes?
Should JLD2 change the header bytes completely to make it more distinct from JLD?

JonasIsensee:
This PR changes the (magic) header bytes to something that is less similar to what JLD uses.

Files with the old header will still be read but a commented out warning message has already been added.
This will become relevant when/if we make changes that break backwards compatibility.

Tests for this will fail since the tests rely on FileIO that does not recognize the new magic bytes yet.

closes #211

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -11,18 +11,13 @@ const OBJECT_HEADER_SIGNATURE = htol(0x5244484f) # "OHDR"
 # Currently we specify that all offsets and lengths are 8 bytes
 const Length = UInt64
 
-# Currently we specify a 512 byte header
-const FILE_HEADER_LENGTH = 512
-const CURRENT_VERSION = v"0.2"
-const REQUIRED_FILE_HEADER = "Julia data file (HDF5), version "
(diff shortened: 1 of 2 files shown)


JonasIsensee:
@lhupe Would you do me a favor and test this locally? (e.g. write some files with a released version and attempt to reload them with this checked out)

lhupe:
seems to work; I can read the old file with the new JLD2 version

PR #217, opened 2020-08-14
Remove FileIO dependency

### LINKED ISSUE #204: Remove unnecessary dependency
Currently FileIO is listed as a dependency.
This is only "needed" to define the FileIO types used for the FileIO interface.

I think we should wrap the FileIO interface definitions into a requires block that only gets loaded when the user executes `using FileIO`.

(the relevant code is at the bottom of `src/loadsave.jl`

lhupe:
I moved all FileIO-dependent function to a new file and wrapped it in an `@require`.
I left `loadtodict!` in `loadsave.jl` because it doesn't depend on FileIO, but it's not used outside the FileIO API, so we could also move it to the new file.

closes #204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,9 @@ matrix:
 notifications:
   email: false
 
+install:
+  - julia -e 'using Pkg; Pkg.add("FileIO")'
+
 # script:
 #  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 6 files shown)


JonasIsensee:
Hm, you removed the FileIO dependency but it probably still needs to remain a test dep?

lhupe:
I have now changed the `.travis.yml` file to install FileIO. Alternatively, one could also add the installation to `runtests.jl`, but I'm not sure which would be "cleaner".

lhupe:
This will hopefully also make code coverage reports (which have been silently broken for two years) work again

PR #221, opened 2020-08-16
An implementation of an interleaved isbits union

JonasIsensee:
This PR is analogous to #212 but pursues a different data layout.

#212 only allows `Vector{Union{T1, T2}} where {T1, T2}` and converts them into a custom struct.

This PR on the other hand implements an element-wise representation for isbits union fields. This has the advantage of using the regular array storage of the HDF5 format.
(h5dump displays your structure which it does not for #212 .

```
"""
    InlineUnionEl{T1,T2}(mask::UInt8, t1::T1, t2::T2)
Custom serialization struct for two member isbits union
fields e.g. in other structs or arrays.
To indicate that t1 is relevant the mask takes the value `UInt8(0)`
and for t2 `UInt8(255)`
"""
struct InlineUnionEl{T1,T2}
    mask::UInt8
    t1::T1
    t2::T2
end
```


JonasIsensee:
```
julia> @time using JLD2
[ Info: Precompiling JLD2 [033835bb-8acc-5ee8-8aae-3f567f8a3819]
  4.090665 seconds (2.15 M allocations: 114.999 MiB, 0.28% gc time)

julia> u = Union{Float64, Missing}[zeros(10^6);];

julia> @time @save "test.jld2" u
  4.864063 seconds (12.90 M allocations: 650.482 MiB, 5.46% gc time)

julia> @time @save "test.jld2" u
  0.063690 seconds (5.20 k allocations: 15.539 MiB)

julia> @time @save "testcompressed.jld2" {compress=true} u
  0.117254 seconds (5.64 k allocations: 32.728 MiB, 5.53% gc time)

julia> using JLD

julia> u = Union{Float64, Missing}[zeros(10^6);];

julia> @time @save "test.jld" u
 61.623914 seconds (18.44 M allocations: 724.206 MiB, 0.41% gc time)

julia> @time @save "test.jld" u
 58.175550 seconds (16.00 M allocations: 602.762 MiB, 0.37% gc time)

julia> @time JLD.jldopen(f->f["u"]=u, "testcompressed.jld", "w", compress=true)
 65.954995 seconds (23.46 M allocations: 973.688 MiB, 0.74% gc time)

julia> using BSON

julia> u = Union{Float64, Missing}[zeros(10^6);];

julia> @time @save "test.jld" u
ERROR: LoadError: UndefVarError: @save not defined
in expression starting at REPL[3]:1

julia> @time BSON.@save "test.bson" u
  1.511677 seconds (11.02 M allocations: 418.530 MiB, 26.47% gc time)

julia> @time using Serialization
  0.000514 seconds (624 allocations: 40.344 KiB)

julia> @time serialize("juliaserializer", u)
  0.488430 seconds (1.91 M allocations: 65.454 MiB, 1.83% gc time)
```

## File sizes
 - Serialization: 8,6M
 - BSON: 16M
 - JLD: 300M / 300 M
 - JLD2: 8,6M / 23K

Sure, we didn't actually store any interesting data here but still.
No one else seems to be able to compress isbits union arrays.

JonasIsensee:
Summary:
In one of the earlier commits I had implemented a `CustomSerialization` that would convert data just as it was written to disk.
This seemed to somewhat work but in general `CustomSerialization` is said to be broken!?

The second problem was that then the data could not be compressed. (because conversion happened too late)
Instead, this PR now converts the data to the inline repr. struct on a relatively high and likely somewhat unintended level.

Thus only `Array`s of two member isbits unions are affected.
if necessary this could be changed again in future releases. For backwards compatibility only the `InlineUnionEl` struct and `read_array` methods would have to remain implemented.

Issue #201, opened 2020-07-18
When should saving fail?


JonasIsensee:
In many cases, when JLD2 fails to save data correctly, the error messages are quite unhelpful.

I plan to improve that but before I would like to discuss, when JLD2 actually *should* fail.

Examples:
 - The filename is not a proper path ( current behaviour: error)
 - MMAP fails to open a file ( current behaviour: segfault - can be triggered by trying to `open` a folder or so)
 - The file already exists (current behaviour: overwritten in mode "w")

One way to improve this would be a better checking of the file path e.g.
 - Is this a well-formed path?
 - Do all the directories exist ? (Should directories be created if not?)
 - Is there a (valid) filename?
 - If the filename is mal-formed, should the data be saved to something like `/tmp/tempname.jld2` to avoid data loss?

If MMAP fails:
 - Use `IOStream` as a fallback instead of erroring ?

What do you think? Would these ideas be valuable improvements or are some of them beyond the scope of JLD2 and should better be implemented elsewhere ?

Eben60:
Saving into `/tmp/` (and producing a warning) is a good idea IMO. If saving into designated path was not possible, but the file name itself was valid, then this name can be used within `/tmp/`

Creation of directories should not be IMO the responsibility of JLD2

Overwriting files (current behavior) may easily result in a serious damage. Acceptable behavior would be to produce another file like foo.0001.jld2, or throw an error / save to tmp. Having an additional argument to enable overwriting could be an option.

JonasIsensee:
I'm actually not sure whether either of these things are good ideas. That's why we need discussion.
What do other libraries do to handle these cases?

Writing to tmp sounds good but imagine the following
```
using JLD2 
for i=1:1000
    #some code
     res = rand(10000000)
    @save "bad/path-$i.jld2" res
end
```
Here one could very well argue that an error in the first iteration would be much better than slowly crashing you computer by filling up ram and then swap space.

Also, this concerns both points: If my code doesn't error, I'd typically assume that my data was saved to the correct place. 
Personally I never interact directly with my data files (e.g. Look at them in the file explorer).

lhupe:
> Use `IOStream` as a fallback instead of erroring ?

I think if a fallback behaviour is implemented, there should be an option to disable it. 
A simple idea for implementing this is to modify `openfile` by adding a `fallback` argument (see below), which could be passed on to the user as a kwarg of `jldopen`
```julia
[...]

PR #225, opened 2020-08-27
Implement backend fallback behaviour

lhupe:
as discussed in #201.
Also ensures that only paths that either refer to a regular file or do not refer to any file system entity at all are passed to `openfile`. This prevents the weird error messages when trying to mmap directories (or even devices, although I haven't tried that yet).

I haven't made sure that the fallback option is accessible from all APIs yet, and I'm going to try to add a few tests as well.

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -183,22 +183,37 @@ h5offset(f::JLDFile, x::Int64) = RelOffset(x - FILE_HEADER_LENGTH)
 # File
 #
 
-openfile(::Type{IOStream}, fname, wr, create, truncate) =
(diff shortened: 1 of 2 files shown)


JonasIsensee:
Your test failed on one of the nightlies. I restarted the test and
on master I just allowed failures on nightly.
If tests still fail, you could rebase on master.

lhupe:
I've added a few simple tests for the fallbacks (and also for the path checking), but I'm not sure they are in the right place. The tests are a bit messy in general, they could do with cleaning up­‌‌‌…

lhupe:
Test fail on Mac when closing an unused `MmapIO`, which seems odd to me

PR #231, opened 2020-09-13
use ftruncate instead of pwrite

### LINKED ISSUE #1: Determine ideal way to enlarge a file
It seems like `pwrite` is faster on Linux and `truncate` is faster on OS X. But perhaps on Linux and other platforms that support it we should use `posix_fallocate`. Also just because `truncate` is faster doesn't mean it's the right approach; it may cause undesirable fragmentation vs. other approaches.

### LINKED ISSUE #55: File saved with JLD2 can't be read, from identical OS and JLD2 version

JonasIsensee:
The specification for `MMAP` says:
> Note that references beyond the end of the object do not extend the object as the new end cannot be determined precisely by most virtual memory hardware. Instead, the size can be directly manipulated by ftruncate().
[https://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html](https://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html)

This seems to be the problem with `pwrite`. It works on most systems but not on NFS.
Instead it is recommended to use `ftruncate`.

Anyone reading this, please compare the speed of writing files  with `JLD2v0.2.0` and this PR.
And if you have encountered #55 before please check whether it is fixed using this PR.

closes #55   !!
closes #1

### DIFF
--- a/src/mmapio.jl
+++ b/src/mmapio.jl
@@ -121,12 +121,14 @@ end
 Base.show(io::IO, ::MmapIO) = print(io, "MmapIO")
 
 if Sys.islinux()
-    # This is substantially faster than truncate on Linux, but slower on OS X.
+    # This is used to be "pwrite" but that is actually not defined behaviour
+    # and fails on NFS systems. "ftruncate" is recommended instead
     # TODO: Benchmark on Windows
     grow(io::IOStream, sz::Integer) =
-        systemerror("pwrite", ccall(:jl_pwrite, Cssize_t,
-                                    (Cint, Ref{UInt8}, UInt, Int64),
-                                    fd(io), UInt8(0), 1, sz - 1) < 1)
+        systemerror("ftruncate",
+            ccall(:ftruncate, Cint, (Cint, Int64),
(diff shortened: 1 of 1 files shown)


JonasIsensee:
Running the benchmarks on my laptop does not show any slowdown.
Real world tests would still be very much appreciated!

PR #234, opened 2020-09-18
Fixes a bug where read would just read up to sizeof(T)

grero:
I came across this bug when looking into why all dataset/group names only contained the first character of the name. It turned out that it was always just reading the number of bytes corresponding to the size of the type, not the request number of bytes.

### DIFF
--- a/src/bufferedio.jl
+++ b/src/bufferedio.jl
@@ -100,7 +100,7 @@ function Base.read(io::BufferedReader, ::Type{T}, n::Int) where T
     buffer = io.buffer
     n = sizeof(T) * n
     if length(buffer) - position < n
-        readmore!(io, sizeof(T))
+        readmore!(io, n)
     end
     io.position[] = position + n
     arr = Vector{T}(undef, n)
--- a/src/bufferedio.jl
+++ b/src/bufferedio.jl
@@ -98,11 +98,11 @@ end
 function Base.read(io::BufferedReader, ::Type{T}, n::Int) where T
     position = io.position[]
     buffer = io.buffer
-    n = sizeof(T) * n
-    if length(buffer) - position < n
-        readmore!(io, n)
+    m = sizeof(T) * n
+    if length(buffer) - position < m
+        readmore!(io, m)
     end
-    io.position[] = position + n
+    io.position[] = position + m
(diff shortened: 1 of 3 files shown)


JonasIsensee:
Hi!
Thanks so much for taking the initiative here.
Do you think you could add a test for this?
Looking at codecov I saw that this line of code actually is not hit in any of the tests.
https://codecov.io/gh/JuliaIO/JLD2.jl/src/master/src/bufferedio.jl

grero:
It must constitute a pretty rare condition, then. Let me see if I can come up with a test. I actually came across it when I copied and modified the bufferedio.jl and dataio.jl files to my own project, which reads JLD2 files over sftp. That probably exposed the bug in ways that reading normal files wouldn't.

grero:
Actually, I think the reason why I hit the bug in my code is that I didn't know how to implement` bytesavailable` for my stream, so i just set it to return 0. That way, the buffer is never grown beyond the immediately requested read size, and the line in question gets triggered. At least that's why I think happened.

JonasIsensee:
That makes sense,
so for testing purposes we could figure out what `bytesavailable` returns for normal IOStreams
and then make it load a string longer than that. (if that is possible - it could be that always all bytes are available)

JonasIsensee:
Alternatively, we could trigger that branch directly
by using `Mocking.jl` https://github.com/invenia/Mocking.jl
and plugging in a faulty `bytesavailable` definition

grero:
I like the idea of using `Mocking.jl`. I'll see if I can use that to produce a test case.

grero:
I added a test for `Base.read(io::BufferedReader, ::Type{T}, n::Int)` by just writing a simple array of `Float64` to a file and checking that I can read back the same array. It hits the missing lines, but we probably want to integrate this better with the other tests.

JonasIsensee:
Hi,
thanks a lot!
I wouldn't worry about integrating it with the other tests. 
Tests have a major restructuring overdue anyway - that is beyond the scope of this PR.

Would you say this is ready for merging ?

grero:
I think this should be OK to merge.

JonasIsensee:
Thanks again!

PR #245, opened 2020-10-10
Make custom serialization possible

JonasIsensee:
This is a (likely still wildly inefficient) attempt at making custom serialization possible in JLD2.

Currently the problem is that the functions that figure out the on disk representation and the associated types are `@generated` functions - namely `h5type`, `jlconvert`, and  `odr`
This works only if the methods for `JLD2.writeas` are defined prior to the `@generated` definition which restricts usage to within JLD2.

In this PR I redefine the above three function definitions to regular functions.

At this point this seems to slow down things significantly. (That is also no surprise as these functions are called very often inside the code - sometimes unnecessarily so - since they were expected to be fast)
Whether this can be sped up to a point where this additional feature becomes an acceptable compromise
remains to be seen.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JLD2"
 uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
-version = "0.2.4"
+version = "0.2.5"
 
 [deps]
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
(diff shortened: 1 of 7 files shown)


JonasIsensee:
there seems to be an unrelated failure in building LIBSVM on windows

JonasIsensee:
Also from running the tests it seems that the runtime is not affected that much.
Compile time is increased (I think)

PR #254, opened 2020-10-28
WIP: Switch to Blosc Compression

### LINKED ISSUE #135: Load times for compressed files grow linearly with size
I noticed that load times for .jld2 files were unreasonably high when compression was used and large arrays were stored, so I made a simple benchmark that compares with JLD.

The good news is that file sizes in JLD2 are only about half as large as in JLD. The bad news ... well, see for yourself.

```julia
using JLD2, FileIO
# using JLD

function test_save_and_reload(n, k)
    x = repeat(rand(n,n), inner=(k,k))  # make fake data that compresses well
    filename = "test.jld2"
    t1 = @elapsed save(filename, "x", x, compress=true)
    t2 = @elapsed load(filename, "x")
    kb = round(Int, filesize(filename)/1024)
    println("(n, k) = ($n, $k), size $(size(x)): save time $(round(t1,digits=2)), load time $(round(t2,digits=2)), file size $kb kB")
end

for n in [100, 200, 500, 1000, 2000, 3000]
    test_save_and_reload(n, 10)
end

for n in [10, 20, 50, 100, 200, 300]
    test_save_and_reload(n, 100)
end
```

Results with JLD, (n, 10) and (n, 100):
```
(n,k) = (100,10), size = (1000, 1000): save time 0.67, load time 0.26, file size 838 kB
(n,k) = (200,10), size = (2000, 2000): save time 0.05, load time 0.04, file size 5543 kB
(n,k) = (500,10), size = (5000, 5000): save time 0.16, load time 0.16, file size 25423 kB
(n,k) = (1000,10), size = (10000, 10000): save time 0.81, load time 0.59, file size 162379 kB
(n,k) = (2000,10), size = (20000, 20000): save time 3.16, load time 2.17, file size 636243 kB
(n,k) = (3000,10), size = (30000, 30000): save time 12.27, load time 8.03, file size 2781257 kB

(n,k) = (10,100), size (1000, 1000): save time 0.01, load time 0.01, file size 108 kB
(n,k) = (20,100), size (2000, 2000): save time 0.03, load time 0.06, file size 656 kB
(n,k) = (50,100), size (5000, 5000): save time 0.09, load time 0.12, file size 2903 kB
(n,k) = (100,100), size (10000, 10000): save time 0.38, load time 0.41, file size 19375 kB
(n,k) = (200,100), size (20000, 20000): save time 1.30, load time 1.28, file size 75436 kB
(n,k) = (300,100), size (30000, 30000): save time 3.21, load time 3.41, file size 305557 kB
```

Results with JLD2, (n, 10) and (n, 100):
```
(n,k) = (100,10), size (1000, 1000): save time 2.33, load time 1.63, file size 135 kB
(n,k) = (200,10), size (2000, 2000): save time 0.43, load time 7.94, file size 549 kB
(n,k) = (500,10), size (5000, 5000): save time 3.14, load time 9.56, file size 20067 kB
(n,k) = (1000,10), size (10000, 10000): save time 12.20, load time 41.78, file size 80444 kB
(n,k) = (2000,10), size (20000, 20000): save time 48.45, load time 181.24, file size 322278 kB
(n,k) = (3000,10), size (30000, 30000): failed (> 10 minutes)

(n,k) = (10,100), size (1000, 1000): save time 2.32, load time 2.23, file size 43 kB
(n,k) = (20,100), size (2000, 2000): save time 0.46, load time 15.05, file size 175 kB
(n,k) = (50,100), size (5000, 5000): save time 2.51, load time 65.87, file size 2318 kB
(n,k) = (100,100), size (10000, 10000): save time 9.75, load time 290.65, file size 9488 kB
(n,k) = (200,100), size (20000, 20000): failed (> 10 minutes)
(n,k) = (300,100), size (30000, 30000): failed (> 10 minutes)
```

JonasIsensee:
This PR makes JLD2 use `Blosc` compression by default.
In principle it is now possible to support different compression algorithms at the same time but so far the API does not allow that.

Initial tests show significant speedups over `CodecZLib`.
The second commit adds additional methods for primitive types that avoids an intermediate step.

closes #135 

TODO: `Blosc` only supports compression of chunks <2GB but `JLD2` has not yet implemented chunking.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 version = "0.2.3"
 
 [deps]
+Blosc = "a74b3585-a348-5f62-a45c-50e91977d574"
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -1,5 +1,6 @@
 module JLD2
-using DataStructures, CodecZlib, Requires
+using DataStructures, CodecZlib, Blosc
+using Requires
 import Base.sizeof
 using MacroTools
 using Printf
@@ -445,6 +446,7 @@ include("loadsave.jl")
 include("stdlib.jl")
 include("backwards_compatibility.jl")
 include("inlineunion.jl")
+include("compression.jl")
 
 function __init__()
     @require FileIO="5789e2e9-d7fb-5bc7-8068-2c6fae9b9549" include("fileio.jl")
--- a/src/compression.jl
+++ b/src/compression.jl
@@ -20,15 +20,34 @@ const DEFLATE_PIPELINE_MESSAGE = let
     take!(io)
 end
 
+const MAX_CHUNK_SIZE = 2^30
+
 function deflate_data(f::JLDFile, data::Array{T}, odr::S, wsession::JLDWriteSession) where {T,S}
-    buf = Vector{UInt8}(undef, odr_sizeof(odr) * length(data))
-    cp = Ptr{Cvoid}(pointer(buf))
-    @simd for i = 1:length(data)
-        @inbounds h5convert!(cp, odr, f, data[i], wsession)
-        cp += odr_sizeof(odr)
+    @show el_size= odr_sizeof(odr)
+    @show num_el = length(data)
+    #num_buffers = (num_el*el_size) ÷ MAX_CHUNK_SIZE + 1 # Approximately
+    buffers = Vector{UInt8}[]
+    idx = 1
+    while idx <= num_el
+        num_el_buffer = min(num_el-idx+1, MAX_CHUNK_SIZE ÷ el_size)
+        buf = Vector{UInt8}(undef, el_size * num_el_buffer)
+        @info "$(el_size*num_el_buffer) bytes in buffer indices $(extrema(range(idx, length=num_el_buffer)))"
+        cp = Ptr{Cvoid}(pointer(buf))
+        @simd for i = range(idx, length=num_el_buffer)
+            @inbounds h5convert!(cp, odr, f, data[i], wsession)
+            cp += el_size
+        end
+        bufc = Blosc.compress(buf; itemsize=el_size)
+        @info "Size after compression $(length(bufc))"
+        push!(buffers, bufc)
+        idx += num_el_buffer
     end
-    Blosc.compress(buf; itemsize=odr_sizeof(odr))
+    buffers
 end
+
+#= function deflate_data(f::JLDFile, data::Array{T}, odr::Type{T}, wsession::JLDWriteSession) where {T,S}
+    Blosc.compress(data)
+end =#
 
 
 @inline function read_compressed_array!(v::Array{T}, f::JLDFile{MmapIO},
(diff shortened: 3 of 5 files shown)


JonasIsensee:
Borrowing code from #135 

```julia
using JLD2, FileIO

function test_save_and_reload(n, k)
    x = repeat(rand(n,n), inner=(k,k))  # make fake data that compresses well
    filename = "test.jld2"
    t1 = @elapsed @save filename {compress=true} x#save(filename, "x", x, compress=true)
    t2 = @elapsed load(filename, "x")
    kb = round(Int, filesize(filename)/1024)
    println("(n, k) = ($n, $k), size $(size(x)): save time $(round(t1,digits=2)), load time $(round(t2,digits=2)), file size $kb kB")
end

for n in [100, 200, 500, 1000]
    test_save_and_reload(n, 10)
end

for n in [10, 20, 50, 100]
    test_save_and_reload(n, 100)
end
```

Gives for `v0.2.4`:
```
(n, k) = (100, 10), size (1000, 1000): save time 1.84, load time 1.47, file size 135 kB
(n, k) = (200, 10), size (2000, 2000): save time 0.24, load time 0.09, file size 549 kB
(n, k) = (500, 10), size (5000, 5000): save time 1.7, load time 0.7, file size 20069 kB
(n, k) = (1000, 10), size (10000, 10000): save time 7.05, load time 4.52, file size 80455 kB
(n, k) = (10, 100), size (1000, 1000): save time 0.09, load time 0.04, file size 43 kB
(n, k) = (20, 100), size (2000, 2000): save time 0.22, load time 0.2, file size 175 kB
(n, k) = (50, 100), size (5000, 5000): save time 1.3, load time 0.53, file size 2317 kB
(n, k) = (100, 100), size (10000, 10000): save time 5.38, load time 3.83, file size 9483 kB
```

And for this PR:
```
(n, k) = (100, 10), size (1000, 1000): save time 1.62, load time 1.41, file size 457 kB
(n, k) = (200, 10), size (2000, 2000): save time 0.03, load time 0.02, file size 1845 kB
(n, k) = (500, 10), size (5000, 5000): save time 0.24, load time 0.2, file size 14695 kB
(n, k) = (1000, 10), size (10000, 10000): save time 0.85, load time 0.56, file size 78667 kB
(n, k) = (10, 100), size (1000, 1000): save time 0.04, load time 0.0, file size 45 kB
(n, k) = (20, 100), size (2000, 2000): save time 0.01, load time 0.03, file size 188 kB
(n, k) = (50, 100), size (5000, 5000): save time 0.05, load time 0.09, file size 1526 kB
(n, k) = (100, 100), size (10000, 10000): save time 0.21, load time 0.42, file size 8440 kB
```

JonasIsensee:
To allow for proper chunking the `chunked_storage_message`
needs to be modified to use `FixedArrayIndex` instead of the `Single Chunk Index`

https://bitbucket.hdfgroup.org/pages/HDFFV/hdf5doc/master/browse/html/H5.format.html#FixedArray

JonasIsensee:
> To allow for proper chunking the `chunked_storage_message`
> needs to be modified to use `FixedArrayIndex` instead of the `Single Chunk Index`
> 
> https://bitbucket.hdfgroup.org/pages/HDFFV/hdf5doc/master/browse/html/H5.format.html#FixedArray

This is hard.
The documentation is not very clear on how to implement this.
What should be the value of `Chunked Storage Property Description` `feature flag` ?
How is `paging` for the `Fixed Array Data Block` turned on and off?

JonasIsensee:
The latest commit adds some code that implements chunking of arrays >1GB with successful reading (and lots of debug print statements).

Annoyingly this is still not compatible to HDF5 because I can't figure out what the docs are trying to tell me.

JonasIsensee:
One approach would be to make the API more flexible and allow the user to choose which compression algorithm he/she wants to use. Then we could just add a warning to the Docs that Blosc can't handle datasets > 2GB and let the user deal with it manually.

I'm open for API suggestions!

I could imagine to add optional arguments to `write` or add an additional signatiure `writecompressed(f, "name", data; compressor)`

JonasIsensee:
To avoid making all compressor libraries dependencies we might be able to dynamically attempt to load them when needed, similar to what is done in `FileIO`.

lhupe:
maybe you could keep the current `compress` kwarg, with `true` and `false` keeping their old behaviour (Zlib or no compression) but also accept symbols to specify alternative algorithms. And if chunking ever matures, you can just tag a new major release and switch the behaviour of `true` from Zlib to Blosc

JonasIsensee:
that might be an option.
Then we would need to maintain a list of compatible compression libraries along with their hdf5 id and symbol.
The default could still be moved to e.g. LZ4 since that is compatible without chunking and is also at least four times faster.

I think it would still be good to widen the API to allow passing an override for specific datasets.
I'd say it is perfectly reasonable that you might want to compress some arrays in a file and not others.
(Of course, you could then also use different compression algs within the same file but even if that's not very useful)

PR #256, opened 2020-10-30
WIP: Initial attempt at adding IO jlread and jlwrite functions.

rofinn:
Figured this probably belongs in JLD2 rather than JLSO. I'm currently stuck on an issue where I can't figure out why writing the objects doesn't put the buffer position at the same place as when reading. Based off an initial example here https://gist.github.com/JonasIsensee/53ad332fd32af92691b40d17a640ead3. Related JLSO issue https://github.com/invenia/JLSO.jl/issues/68

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,0 +1,100 @@
+function jlread(io::IO)
+    startpos = position(io)
+    f = JLDFile(io, "", false, false, false, false)
+
+    @show startpos
+
+    # Extract the file header
+    header = String(read!(io, Vector{UInt8}(undef, length(REQUIRED_FILE_HEADER))))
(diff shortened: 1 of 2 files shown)


src/io.jl
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    @show position(io)
+
+    result = Dict{String, Any}()
+    loadtodict!(result, f)
+    return result
+end
+
+function jlwrite(io::IO, data::AbstractDict)
```
rofinn, line 59: This function currently only works correctly with `IOStream`. I feel like the core logic of JLD2 is geared to heavily for file io streams for this to be a useful PR.

Issue #258, opened 2020-11-17
Saving and loading (anonymous) functions


hendri54:
When saving and then loading a `struct` that contains a function I now get a MethodError. MWE:

```
using FileIO, JLD2

f1() = 1.0;

mutable struct Foo
    x :: Float64
    f :: Function
end

x = Foo(1.0, f1);

fPath = "test1.jld2";
save(fPath, Dict("data" => x));

y = load(fPath)
```

The error is:

```
ERROR: LoadError: MethodError: no method matching typeof(f1)()
Stacktrace:
 [1] handle_error(::MethodError, ::File{DataFormat{:JLD2}}) at /Users/lutz/.julia/packages/FileIO/2fEu2/src/error_handling.jl:82
 [2] handle_exceptions(::Array{Any,1}, ::String) at /Users/lutz/.julia/packages/FileIO/2fEu2/src/error_handling.jl:77
 [3] load(::Formatted; options::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{,Tuple{}}}) at /Users/lutz/.julia/packages/FileIO/2fEu2/src/loadsave.jl:210
 [4] load at /Users/lutz/.julia/packages/FileIO/2fEu2/src/loadsave.jl:187 [inlined]
 [5] #load#16 at /Users/lutz/.julia/packages/FileIO/2fEu2/src/loadsave.jl:136 [inlined]
 [6] load(::String) at /Users/lutz/.julia/packages/FileIO/2fEu2/src/loadsave.jl:136
 [7] top-level scope at /Users/lutz/Documents/projects/p2019/college_stratification/CollegeStrat/src/temp1.jl:15
 [8] include(::String) at ./client.jl:457
 [9] top-level scope at REPL[9]:1
in expression starting at /Users/lutz/Documents/projects/p2019/college_stratification/CollegeStrat/src/temp1.jl:15
```

Replacing the function with, say, and `Int` avoids the error.

This is on Julia 1.5.2, JLD2 0.3.0, FileIO 1.4.4.

JonasIsensee:
Hi @hendri54 ,

thanks for this bug report!
The simple example made the problem very clear.

The problem was introduced in `v0.3.0` and, until this is fixed, it might be best if you continue working with JLD2 `v0.2.4`.

The problem is in 
https://github.com/JuliaIO/JLD2.jl/blob/d6bd50889667312d90b5c80b830bb797354706a7/src/data.jl#L1339

[...]

PR #260, opened 2020-11-18
use new instead of T()

JonasIsensee:
`v0.3.0` introduced a bug in that functions can no longer be reconstructed. (#258)

The offending thing was essentially to try `typeof(f)()` to get the function from its type.

With this change it seems to work but I have no absolutely no idea whether this is a good idea or not.

Any ideas or recommendations ?


JonasIsensee:
Since this approach is already in use inside BSON and it passes all existing plus new tests
I will merge this now.

Issue #281, opened 2021-02-19
Error when loading a dict saved on a 64-bits machine in a 32-bit one, and vice-versa


carlobaldassi:
I have encountered an error when trying to load a `Dict` created with a 64-bit machine in a 32-bit one. It only happens with some value types, but it also happens in the opposite direction (32 -> 64). This is the most reduced version that I could find that reproduces the problem:

```
# on a 64-bit machine do this
julia> d = Dict{String,Vector}("v"=>[0.0]) # same problem with Vector{Float64}, not with Array nor Any
julia> @save "d.jld2" d
```

```
# then transfer the file on a 32-bit machine and load it:
julia> @load "d.jld2" d
ERROR: MethodError: no method matching Array{Float64, 1}(::Vector{Float64})
Closest candidates are:
  Array{T, N}(::AbstractArray{S, N}) where {T, N, S} at array.jl:540
  Array{T, N}(::BitArray{N}) where {T, N} at bitarray.jl:494
  Array{T, N}(::Core.Compiler.BitArray{N}) where {T, N} at bitarray.jl:494
  ...
Stacktrace:
  [1] convert(#unused#::Type{Array{Float64, 1}}, a::Vector{Float64})
    @ Base ./array.jl:532
  [2] jlconvert
    @ ~/.julia/packages/JLD2/HYDlD/src/data.jl:686 [inlined]
  [3] macro expansion
    @ ~/.julia/packages/JLD2/HYDlD/src/data.jl:1433 [inlined]
  [4] jlconvert(#unused#::JLD2.ReadRepresentation{Pair{String, Array{Float64, 1}}, JLD2.OnDiskRepresentation{(0, 16), Tuple{String, Array{Float64, 1}}, Tuple{JLD2.Vlen{String}, JLD2.RelOffset}}()}, f::JLD2.JLDFile{JLD2.MmapIO}, ptr::Ptr{Nothing}, header_offset::JLD2.RelOffset)
    @ JLD2 ~/.julia/packages/JLD2/HYDlD/src/data.jl:1378
  [5] read_scalar
    @ ~/.julia/packages/JLD2/HYDlD/src/dataio.jl:37 [inlined]
[...]

PR #283, opened 2021-02-24
WIP: Cross-platform (precision) type reconstruction

JonasIsensee:
Convert any type parameter that is `<: Union{Int64, Int32}` to the system's integer precision `Int`
prior to instantiating a parametric data type.
Adresses #281


carlobaldassi:
I tested with simple examples like the ones in #281 and it seems to be working, in both directions.

PR #285, opened 2021-02-26
WIP: Allow custom expected group size

JonasIsensee:
Adds two additional keyword arguments to constructors of `Group` ,
namely `est_num_entries == 4` and `est_link_name_len == 8` to allow 
tune how much space should be left at the end of a group table
for future appending.

(Also implements appending when there is space left)
These two settings are stored inside group metadata and are recovered when loading a file.
They are also used to compute how much additional space should be allocated when the existing space runs out.

`est_num_entries` specifies how many entries the group is expected to contain
`est_link_name_len` allows to further tune that by specifying the expected average number of bytes per dataset `name`

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -116,18 +116,26 @@ mutable struct Group{T}
     last_chunk_start_offset::Int64
     continuation_message_goes_here::Int64
     last_chunk_checksum_offset::Int64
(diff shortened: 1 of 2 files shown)


JonasIsensee:
Any ideas for better kwarg names?

PR #297, opened 2021-04-06
jldsave function API

JonasIsensee:
This PR adds a new API function that leverages julia's keyword argument syntax.

```
julia> using JLD2
[ Info: Precompiling JLD2 [033835bb-8acc-5ee8-8aae-3f567f8a3819]

julia> hello = 1
1

julia> jldsave("test.jld2"; hello, world = 2)

julia> using FileIO

julia> load("test.jld2")
Dict{String, Any} with 2 entries:
  "hello" => 1
  "world" => 2
```


lhupe:
hmm, I like this syntax a lot, but I'm starting to think that it might be more useful to add it to the FileIO API (as an alternative zero-argument method). After all, it does basically the same thing (store key-value pairs in a JLD2 file) and I personally think the current `FileIO.save("filename.jld2", "key1", value1, "key2", value2)` syntax is a bit weird.

johnnychen94:
A compatible workaround to https://github.com/JuliaIO/FileIO.jl/issues/324:

```julia
fileio_save(f::File{format"JLD2"}, ::Nothing; kwargs...) = jldsave(FileIO.filename(f); kwargs...)
```

and use `save("test.jld2", nothing; x, y, zebra=z)`. How do you think of this?

JonasIsensee:
> A compatible workaround to [JuliaIO/FileIO.jl#324](https://github.com/JuliaIO/FileIO.jl/issues/324):
> 
> ```julia
> fileio_save(f::File{format"JLD2"}, ::Nothing; kwargs...) = jldsave(FileIO.filename(f); kwargs...)
> ```
> 
> and use `save("test.jld2", nothing; x, y, zebra=z)`. A little bit weird for `nothing` though but it works. How do you think of this?

Hm, no. 
I don't think that is a valid option. I'd rather use a new function name than having to pass an arbitrary `nothing` just to make dispatch work.

johnnychen94:
https://github.com/JuliaIO/FileIO.jl/issues/324 is a breaking change so even if it's approved, it needs time to adopt. Before that actually get supported, `save("test.jld2", nothing; x, y, zebra=z)` is a forward-compatible workaround.

PR #306, opened 2021-04-28
[doc] add some more information about compression.

fredrikekre:


JonasIsensee:
Thank you, that is a good addition!

PR #310, opened 2021-05-02
inline group show

JonasIsensee:
@lhupe Is this correct?

It fixed it for me, locally


lhupe:
The code looks fine, but I'm not sure what you mean by “it”.

PR #317, opened 2021-05-18
Remove Requires, as appears to be unused?

IanButterworth:
I searched for `@require`'s but couldn't find any. Seems like a dep that can be dropped?


JonasIsensee:
Hi @IanButterworth ,
good catch! Thank you :)

IanButterworth:
Great. Could a release be made?

JonasIsensee:
Done. v0.4.6 should be out soon

PR #324, opened 2021-06-05
ImmutableDict serialization

YingboMa:


JonasIsensee:
Hi @YingboMa ,

thanks! This appears to be a very sensible addition.
Out of curiosity: How did you notice that it wasn't already implemented?

There appears to be an error in the tests for julia `v1.0` that's related to the `ImmutableDict` constructor.

ChrisRackauckas:
>Out of curiosity: How did you notice that it wasn't already implemented?

Failures when being used with Symbolics.jl

ChrisRackauckas:
>There appears to be an error in the tests for julia v1.0 that's related to the ImmutableDict constructor.

Just drop v1.0?

JonasIsensee:
Nah, some poor souls appear to still be stuck on v1.0. 
There's no need to break compatibility just for this.
(In particular since the problem is only in the tests)

I pushed a change that should fix it.
Will merge once tests pass.

PR #331, opened 2021-06-25
fix #327 DataType has no field ninitialized

### LINKED ISSUE #327: Error `DataType has no field ninitialized` when saving Tuple
```julia
julia> import JLD2

julia> JLD2.save("tst.jld2", "x", (1,2))
Error encountered while save FileIO.File{FileIO.DataFormat{:JLD2}, String}("tst.jld2").

Fatal error:
ERROR: type DataType has no field ninitialized
...
```

 [033835bb] JLD2 v0.4.9

Julia: Version 1.7.0-DEV.1290, 6e8754337a*

EDIT:
The bug does *not* occur in
Julia 1.7.0-DEV.961, 2435f96f92*

jw3126:
fix #327

### DIFF
--- a/src/backwards_compatibility.jl
+++ b/src/backwards_compatibility.jl
@@ -36,7 +36,7 @@ end
     end
 end
 
-@static if :ninitialized in fieldnames(Tuple{})
+@static if :ninitialized in fieldnames(DataType)
     # https://github.com/JuliaIO/JLD2.jl/issues/327
     function ninitialized(@nospecialize(T::Type))::Int
         T.ninitialized
--- a/src/backwards_compatibility.jl
+++ b/src/backwards_compatibility.jl
@@ -36,7 +36,7 @@ end
     end
 end
 
-@static if :ninitialized in fieldnames(DataType)
+if :ninitialized in fieldnames(DataType)
     # https://github.com/JuliaIO/JLD2.jl/issues/327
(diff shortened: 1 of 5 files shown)


src/backwards_compatibility.jl
```diff
@@ -35,3 +35,14 @@ end
         return isa(t, DataType) && t.mutable
     end
 end
+
+@static if :ninitialized in fieldnames(Tuple{})
```
jw3126, line 39: I don't think this needs to be `@static`, I just followed the `@static if VERSION < v"1.7.0-A"` code block above.

src/backwards_compatibility.jl
```diff
@@ -35,3 +35,14 @@ end
...
+    # https://github.com/JuliaIO/JLD2.jl/issues/327
+    function ninitialized(@nospecialize(T::Type))::Int
+        T.ninitialized
+    end
+else
+    function ninitialized(@nospecialize(T::Type))::Int
```
jw3126, line 45: Not sure if there is a benefit of `@nospecialize` here, again I just followed the tradition `function ismutabletype(@nospecialize(t::Type))`

JonasIsensee:
Hi @jw3126 ,
thanks a lot for taking care of this!

jw3126:
@JonasIsensee thanks a lot for maintaining this essential package!

PR #340, opened 2021-07-28
Add precompiles to reduce time to first `jldopen`/`jldsave`

thchr:
In order to reduce the "time-to-first-`jldopen`/`jldsave`" I added a `src/precompile.jl` file populated by SnoopCompile.

Before:
```jl
julia> @time using JLD2
  0.839626 seconds (1.12 M allocations: 72.258 MiB, 2.57% gc time)

julia> @time jldsave("example.jld2"; a=1, b=2, c=[1.0,2.0], d=Dict(s => 2 for s in ["a", "b", "c"]))
  6.407678 seconds (16.59 M allocations: 958.140 MiB, 7.07% gc time, 0.82% compilation time)

julia> @time jldopen("example.jld2") do _ end # fresh session
  2.483562 seconds (7.79 M allocations: 451.146 MiB, 12.52% gc time)
```
After:
```jl
julia> @time using JLD2
  0.997764 seconds (1.41 M allocations: 92.667 MiB, 2.02% gc time)

julia> @time jldsave("example.jld2"; a=1, b=2, c=[1.0,2.0], d=Dict(s => 2 for s in ["a", "b", "c"]))
  4.672978 seconds (6.89 M allocations: 407.284 MiB, 5.46% gc time, 1.50% compilation time)

julia> @time jldopen("example.jld2") do _ end # fresh session
  1.352138 seconds (1.44 M allocations: 86.312 MiB, 3.25% gc time, 99.97% compilation time)
 ```
 
 I.e., a small increase in `using` time and a significant reduction in first load/save time. That seems worthwhile to me?
 
 The precompile statements were added by doing
 ```jl
 using SnoopCompile, Test, FileIO
tinf = @snoopi_deep begin include("test/rw.jl"); include("test/loadsave.jl"); end
ttot, pcs = ttot, pcs = SnoopCompile.parcel(tinf; tmin=0.05)
SnoopCompile.write("tmp", pcs)
 ```
I didn't think terribly carefully about which files to snoop on - so feel free to let me know if other files would be better. I wanted to just do check `runtest.jl` but had issues making that run without erroring out (some issues with dependencies).
Similarly, the `tmin` cutoff for `SnoopCompile.parcel` could be tuned; I picked one that seemed to account for a good chunk of the inference time without having a too-crazy amount of precompile statements.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -11,91 +11,91 @@ function _precompile_()
     ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
 
     Base.precompile(Tuple{typeof(write),JLDFile{MmapIO},String,Float64})   # time: 0.8394732
-    Base.precompile(Tuple{typeof(write),JLDFile{MmapIO},String,Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}}})   # time: 0.5689673
+    Base.precompile(Tuple{typeof(write),JLDFile{MmapIO},String,Base.ReshapedArray{Int, 2, UnitRange{Int}, Tuple{}}})   # time: 0.5689673
     Base.precompile(Tuple{Core.kwftype(typeof(jldopen)),NamedTuple{(:compress,), Tuple{Bool}},typeof(jldopen),String,Bool,Bool,Bool,Type{MmapIO}})   # time: 0.3199544
-    Base.precompile(Tuple{Core.kwftype(typeof(write)),NamedTuple{(:compress,), Tuple{Nothing}},typeof(write),Group{JLDFile{MmapIO}},String,Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}},JLDWriteSession{Dict{UInt64, RelOffset}}})   # time: 0.2051841
-    Base.precompile(Tuple{typeof(read_data),JLDFile{MmapIO},ReadRepresentation{ComplexF64, ComplexF64},Tuple{ReadDataspace, RelOffset, Int64, UInt16},Vector{ReadAttribute}})   # time: 0.187065
-    Base.precompile(Tuple{typeof(read_data),JLDFile{MmapIO},ReadRepresentation{BitMatrix, OnDiskRepresentation{(0, 8, 16), Tuple{Vector{UInt64}, Int64, Tuple{Int64, Int64}}, Tuple{RelOffset, Int64, Tuple{Int64, Int64}}}()},Tuple{ReadDataspace, RelOffset, Int64, UInt16},Vector{ReadAttribute}})   # time: 0.1830797
-    Base.precompile(Tuple{Core.kwftype(typeof(write)),NamedTuple{(:compress,), Tuple{Nothing}},typeof(write),Group{JLDFile{IOStream}},String,Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}},JLDWriteSession{Dict{UInt64, RelOffset}}})   # time: 0.172313
+    Base.precompile(Tuple{Core.kwftype(typeof(write)),NamedTuple{(:compress,), Tuple{Nothing}},typeof(write),Group{JLDFile{MmapIO}},String,Base.ReshapedArray{Int, 2, UnitRange{Int}, Tuple{}},JLDWriteSession{Dict{UInt, RelOffset}}})   # time: 0.2051841
+    Base.precompile(Tuple{typeof(read_data),JLDFile{MmapIO},ReadRepresentation{ComplexF64, ComplexF64},Tuple{ReadDataspace, RelOffset, Int, UInt16},Vector{ReadAttribute}})   # time: 0.187065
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/precompile.jl [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+# spent in executing /test/rw.jl and /test/loadsave.jl
+# A single CodecZlib argument call was filtered out manually.
+
+function _precompile_()
+    ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
+
+    Base.precompile(Tuple{typeof(write),JLDFile{MmapIO},String,Float64})   # time: 0.8394732
+    Base.precompile(Tuple{typeof(write),JLDFile{MmapIO},String,Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}}})   # time: 0.5689673
+    Base.precompile(Tuple{Core.kwftype(typeof(jldopen)),NamedTuple{(:compress,), Tuple{Bool}},typeof(jldopen),String,Bool,Bool,Bool,Type{MmapIO}})   # time: 0.3199544
+    Base.precompile(Tuple{Core.kwftype(typeof(write)),NamedTuple{(:compress,), Tuple{Nothing}},typeof(write),Group{JLDFile{MmapIO}},String,Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}},JLDWriteSession{Dict{UInt64, RelOffset}}})   # time: 0.2051841
```
timholy, line 16: Looks like this package tests 32bit as well as 64bit, and sadly SnoopCompile doesn't handle this (sorry). What about just pasting your test script in after the `ccall` in this function? It should be robust, as long as the side-effect of writing a file is not problematic. Alternatively, replace `Int64` with `Int` in various places?

timholy (commented):
Very nice! Thanks for doing this!

src/precompile.jl
thchr: Ah, good point! Since the test files I ran to generate these need CodecZlib (which is not a direct dependency of JLD2) I think your second suggestion would be easier - so I just did that (and swapped `UInt64` as well).

PS. This PR is a direct consequence of wanting to try out what I learned from yours and Shuhei's recent workshop - so many thanks from me as well.

JonasIsensee:
Thanks a lot @thchr  !
I'd played with this in the past but never got around to finishing it.

src/precompile.jl
timholy: Ah yes, now that I look more carefully you're not saving a single file, you're running tests, and that takes long enough that it's better to write out the precompile directives.

JonasIsensee:
Hi @thchr ,

someone just noticed #341 (and see #342).
What was the reason behind the original version bound you set?

Either way, it's not a big deal, since we don't have to add optimizations to past julia versions anyway.

thchr:
Hmm, the v1.4.2 bound came from SnoopCompile itself (see also e.g. [this part of it's documentation](https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_parcel/#SnoopCompile.write)). I think the precompiled statement basically was not worthwhile before v1.4.2.

I don't immediately understand why this would produce an error on v1.5 though, unfortunately.

JonasIsensee:
I checked out the line with the error.
I think, maybe the type signature of `Base.ReshapedArray` or so may have changed.
(it's not exported so that could be)

thchr:
Ah, yes, I was just about to write that it seems like the OP's precompilation step works fine until it hits this line: 

https://github.com/JuliaIO/JLD2.jl/blob/59b384c58029817fbf1a19d0598bb5c6d133512b/src/precompile.jl#L45

Indeed, it is probably just related to the signature having changed relative to earlier versions of Julia.

timholy:
Yes, if the signature changes between Julia versions, that's not so good. In this case `ReshapedArray` changed its parameters in 1.6.

There may be a safer way to get version-independent precompiles (again by executing code rather than issuing PC statements), so if needed options include (1) commenting out specific problematic lines or (2) reverting for now and try again in a bit. @tchr, it occurs to ask, did you check specializations with `pgdsgui`? If not, it might be worth doing that first, since adding judicious `@nospecialize` can change your signatures and is often much more impact than adding `precompile` directives.

timholy:
@thchr, at the hackathon I'm willing to do some live-video support for latency (see the discord channel). If you want we can try to get together.

thchr:
> @thchr, at the hackathon I'm willing to do some live-video support for latency (see the discord channel). If you want we can try to get together.

@timholy Ah, that's a great offer - I will try to catch you!

timholy:
@thchr, see #344. After disabling the precompile file (which hides things from inference, which is not what you want when running diagnostics), here's what I did:
```julia
julia> using JLD2, SnoopCompile, MethodAnalysis, Profile

julia> Profile.init(delay=0.01, n=10^7)

shell> cd test
/home/tim/.julia/dev/JLD2/test

julia> tinf = @snoopi_deep @profile begin
           include("rw.jl")
           include("loadsave.jl")
       end
<output suppressed>

julia> using PyPlot: PyPlot, plt

julia> mref, ax = pgdsgui(tinf; by=exclusive);
```

I got this:
image

The dots along the left side have essentially no runtime, yet some of them are very expensive in terms of compile time.

From there, you start clicking on the dots and doing things like `edit(mref[])` to see which methods are the most egregious offenders. You can also use `methodinstances` to figure out how many instances there are, or combine `inference_triggers` with `aggregate_by_source(Method, itrigs)` to figure out which methods have the most specializations (though this is indicated by the color of the dot).

Issue #367, opened 2022-01-13
Issue saving large amounts of data


BoZenKhaa:
When trying to save large amount of data, I get
```julia
Fatal error:
ERROR: LoadError: InexactError: trunc(UInt16, 80540)

Stacktrace:                                                                                                 
  [1] throw_inexacterror(f::Symbol, #unused#::Type{UInt16}, val::Int64)                                     
    @ Core ./boot.jl:612                                                                                    
  [2] checked_trunc_uint                                                                                    
    @ ./boot.jl:642 [inlined]                                                                               
  [3] toUInt16                                                                                                           
    @ ./boot.jl:715 [inlined]
  [4] UInt16
    @ ./boot.jl:765 [inlined]
  [5] convert
    @ ./number.jl:7 [inlined]
  [6] JLD2.HeaderMessage(msg_type::UInt8, size::Int64, flags::Int64)                                        
    @ JLD2 ~/.julia/packages/JLD2/DcnTD/src/object_headers.jl:63
  [7] commit(f::JLD2.JLDFile{JLD2.MmapIO}, dt::JLD2.CompoundDatatype, attrs::Tuple{JLD2.WrittenAttribute{JLD2
.WriteDataspace{0, Tuple{}}, JLD2.CommittedDatatype, DataType}, JLD2.WrittenAttribute{JLD2.WriteDataspace{1, 
Tuple{}}, JLD2.BasicDatatype, Vector{JLD2.RelOffset}}})
#.....
```

It seems the failure is caused by `Header.size` being limited to `UInt16`. Is this a hard limitation?

JonasIsensee:
Hi @BoZenKhaa, 

can you give me more information on what you were trying to store?
Maybe we can find a workaround.

From the stacktrace it looks like you were trying to store a very complex struct. 
Also #343

BoZenKhaa:
Hello @JonasIsensee, thank you for looking into this, you are right, after investigating further, the data being stored is below 100MB, but it has a bit of structure to it. 

I am struggling to produce workable MWE, since I got the data by saving larger and larger instances of same stuff, and I had the first failure after couple of iterations. 

The data is generally a dictionary that looks like this, saved through `DrWatson.jl` as `Dict("data"=>dict_below)`:
```
Dict{Symbol, Any} with 2 entries:
[...]

Issue #343, opened 2021-07-29
Saving fails with InexactError on ODE solution with long type name


bgroenks96:
Error:

```
ERROR: LoadError: InexactError: trunc(UInt16, 108752)
Stacktrace:
  [1] throw_inexacterror(f::Symbol, #unused#::Type{UInt16}, val::Int64)
    @ Core ./boot.jl:602
  [2] checked_trunc_uint
    @ ./boot.jl:632 [inlined]
  [3] toUInt16
    @ ./boot.jl:705 [inlined]
  [4] UInt16
    @ ./boot.jl:755 [inlined]
  [5] convert
    @ ./number.jl:7 [inlined]
  [6] JLD2.HeaderMessage(msg_type::UInt8, size::Int64, flags::Int64)
    @ JLD2 ~/.julia/packages/JLD2/m5olh/src/object_headers.jl:63
  [7] commit(f::JLD2.JLDFile{IOStream}, dt::JLD2.CompoundDatatype, attrs::Tuple{JLD2.WrittenAttribute{JLD2.WriteDataspace{0, Tuple{}}, JLD2.CommittedDatatype, DataType}, JLD2.WrittenAttribute{JLD2.WriteDataspace{1, Tuple{}}, JLD2.BasicDatatype, Vector{JLD2.RelOffset}}})
    @ JLD2 ~/.julia/packages/JLD2/m5olh/src/datatypes.jl:277
  [8] commit(f::JLD2.JLDFile{IOStream}, dtype::JLD2.CompoundDatatype, writeas::DataType, readas::DataType, attributes::JLD2.WrittenAttribute{JLD2.WriteDataspace{1, Tuple{}}, JLD2.BasicDatatype, Vector{JLD2.RelOffset}})
    @ JLD2 ~/.julia/packages/JLD2/m5olh/src/data/writing_datatypes.jl:209
  [9] commit_compound(f::JLD2.JLDFile{IOStream}, names::Vector{Symbol}, writtenas::DataType, readas::Type)
    @ JLD2 ~/.julia/packages/JLD2/m5olh/src/data/writing_datatypes.jl:173
 [10] h5type(f::JLD2.JLDFile{IOStream}, writtenas::Type, x::ODESolution{...
```

The type signature of `ODESolution{...}` and my corresponding problem/function is very, very long, so perhaps that has something to do with it? I kind of doubt it's the object size. It's not *that* big.

```
Julia Version 1.6.0
[...]

PR #370, opened 2022-01-27
don't inline long ntuples

JonasIsensee:


JonasIsensee:
hopefull fixes #367 #343

PR #372, opened 2022-01-28
Add a comment on `jldsave`

BoundaryValueProblems:
Emphasize the importance of `;` in the description of `jldsave` function in README.md.
If one simply uses `,` instead of `;` at the start of the list of the variables to be saved, one gets errors.


JonasIsensee:
Thank you! :)

PR #373, opened 2022-01-31
TTFX improvements

JonasIsensee:


JonasIsensee:
So with this tests run a bit faster than before.

More importantly:
on master
```
julia> @time using JLD2
  0.321425 seconds (454.71 k allocations: 31.983 MiB, 6.80% compilation time)

julia> @time begin 
       f=jldopen("test.jld2", "w")
       write(f, "a", 1)
       close(f)
       end
  3.096658 seconds (6.68 M allocations: 345.970 MiB, 3.80% gc time, 99.90% compilation time)

## new session 

julia> @time load("test.jld2")
  4.344244 seconds (8.86 M allocations: 459.389 MiB, 7.50% gc time, 99.97% compilation time)
  "a" => 1
```

with this PR


```
julia> @time using JLD2
  0.472536 seconds (721.30 k allocations: 49.075 MiB, 4.54% compilation time)

julia> @time begin 
       f=jldopen("test.jld2", "w")
       write(f, "a", 1)
       close(f)
       end
  1.557660 seconds (1.99 M allocations: 105.382 MiB, 2.42% gc time, 99.75% compilation time)

## new session

julia> @time load("test.jld2")
  2.823994 seconds (5.45 M allocations: 286.137 MiB, 4.04% gc time, 99.97% compilation time)
  "a" => 1
```

JonasIsensee:
My impression was that JLD2 of the remaining load / compile times much is either not originally from JLD2 or down to code I don't currently want to despecialize.

I'll merge this for now and will maybe pick it up again at some point in the future.

Issue #57, opened 2018-01-19
Allow writing and reading from a stream


iamed2:
Sometimes you want to write to a stream and not a file

iamed2:
I'm working on a PR for this

iamed2:
I wrote something I thought would work but it ended up producing corrupt data. I don't think I know enough about HDF5 to debug. I still think this would be a good feature.

marosko89:
Yes, good idea :)

PR #377, opened 2022-02-08
DRAFT: env + binary blobs

JonasIsensee:
This is a demonstration implementation of two things inspired by `JLSO.jl`

It adds definition `write_env(f)`, `has_env(f)`, `read_env(f)` and `Pkg.activate(f::JLDFile)` to write, check, read, and activate an environment stored in a file.

It also adds a proof of concept custom serialization `Blob`. `Blob` has a single field of type `Any` that gets serialized using `Serialization.serialize`. 

```
julia> f = jldopen("test2.jld2", "w")
JLDFile /home/jonas/.julia/dev/JLD2/test2.jld2 (read/write)
  (no datasets)

julia> JLD2.write_blob(f, "h", ()->rand(2,2))

julia> close(f)

## new session

julia> d = load("test2.jld2")
  "h" => Blob(#7)

julia> d["h"].obj()
2×2 Matrix{Float64}:
 0.983637  0.0874438
 0.641339  0.81356
```

alternatively write `jldsave("test.jld2"; a=JLD2.Blob(()->rand(3,3)))`

a proper API for wrapping and unwrapping in `Blob`s could of course be built.

EDIT: it might be better to store the enviroment metadata as attributes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,7 @@ MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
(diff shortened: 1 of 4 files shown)


JonasIsensee:
Hey @oxinabox,
your ridiculously fast response to me closing issue #57 inspired me to give this some thought again.
I'm still not sure what the best option is  or whether there's a good way to integrate JLD2/HDF5 as an "object" format in JLSO.

However, I noticed that I can actually add some of the nice features of JLSO to JLD2 with fairly little effort.
Just pinging for your thoughts / opinion

oxinabox:
I think this is a nice idea

PR #423, opened 2022-10-13
make compress doc more visible

xgdgsc:
I would consider making compress enabled the default behavior.


JonasIsensee:
thank you for this improvement.

I have considered it in the past but don't think it would be a good idea as default.
Compression requires an additional dependency and is not always useful.
Depending on the data it can even be counter-productive.

PR #435, opened 2022-11-09
avoid spurious bug with eager gc

JonasIsensee:


oameye:
When will this be merged?

PR #437, opened 2022-11-11
Don't access `.size` directly

vchuravy:
This broke on Julia nightly

https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/f1ae425_vs_317211a/Bio3DView.against.log

x-ref: https://github.com/JuliaLang/julia/pull/47170


JonasIsensee:
This doesn't quite yet fix it, since apparently JLD2 sometimes accessed `.size` of objects like `Symbol` which don't have a fixed size. `.size` is then set to `0` but `sizeof` returns an error.

JonasIsensee:
Thanks. This was fixed in #438

Issue #428, opened 2022-10-31
Broken with Julia master


eschnett:
The current master version of JLD2 fails on the current master version of Julia:
```
$ julia +dev
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-DEV.1696 (2022-10-31)
 _/ |\__'_|_|_|\__'_|  |  Commit f70b5e4767* (0 days old master)
|__/                   |

julia> using JLD2
[ Info: Precompiling JLD2 [033835bb-8acc-5ee8-8aae-3f567f8a3819]
ERROR: LoadError: type DataType has no field size
Stacktrace:
 [1] getproperty(x::Type, f::Symbol)
   @ Base ./Base.jl:32
 [2] top-level scope
   @ ~/.julia/dev/JLD2/src/data/number_types.jl:14
 [3] include(mod::Module, _path::String)
   @ Base ./Base.jl:450
 [4] include(x::String)
   @ JLD2 ~/.julia/dev/JLD2/src/JLD2.jl:1
 [5] top-level scope
   @ ~/.julia/dev/JLD2/src/JLD2.jl:577
 [6] include
   @ ./Base.jl:450 [inlined]
 [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
   @ Base ./loading.jl:1667
 [8] top-level scope
   @ stdin:1
in expression starting at /Users/eschnett/.julia/dev/JLD2/src/data/number_types.jl:13
in expression starting at /Users/eschnett/.julia/dev/JLD2/src/JLD2.jl:1
in expression starting at stdin:1
ERROR: Failed to precompile JLD2 [033835bb-8acc-5ee8-8aae-3f567f8a3819] to /Users/eschnett/.julia/compiled/v1.9/JLD2/jl_1Zc39q.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, keep_loaded_modules::Bool)
[...]

PR #438, opened 2022-11-12
Do not access .size field in DataType

eschnett:
Closes https://github.com/JuliaIO/JLD2.jl/issues/428.

### DIFF
--- a/src/data/number_types.jl
+++ b/src/data/number_types.jl
@@ -3,20 +3,19 @@
 # representations than ordinary opaque types
 
 # This construction prevents these methods from getting called on type unions
-const SignedTypes        = Union{Type{Int8}, Type{Int16}, Type{Int32}, Type{Int64}, Type{Int128}}
-const UnsignedTypes      = Union{Type{UInt8}, Type{UInt16}, Type{UInt32}, Type{UInt64}, Type{UInt128}}
-const FloatTypes         = Union{Type{Float16}, Type{Float32}, Type{Float64}}
-const PrimitiveTypeTypes = Union{SignedTypes, UnsignedTypes, FloatTypes, Type{Bool}}
-const PrimitiveTypes     = Union{Bool, Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32,
-                                 UInt64, UInt128, Float16, Float32, Float64}
+const SignedTypes        = [Int8, Int16, Int32, Int64, Int128]
+const UnsignedTypes      = [UInt8, UInt16, UInt32, UInt64, UInt128]
+const FloatTypes         = [Float16, Float32, Float64]
+const PrimitiveTypes     = [SignedTypes; UnsignedTypes; FloatTypes; Bool]
+const PrimitiveTypeTypes = Union{(Type{T} for T in PrimitiveTypes)...}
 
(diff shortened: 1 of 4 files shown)


eschnett:
The failing check is a spurious small apparent reduction in coverage because codecov miscounts some function metadata. There is no actual reduction in coverage.

src/data/number_types.jl
```diff
@@ -3,20 +3,19 @@
...
-const UnsignedTypes      = Union{Type{UInt8}, Type{UInt16}, Type{UInt32}, Type{UInt64}, Type{UInt128}}
-const FloatTypes         = Union{Type{Float16}, Type{Float32}, Type{Float64}}
-const PrimitiveTypeTypes = Union{SignedTypes, UnsignedTypes, FloatTypes, Type{Bool}}
-const PrimitiveTypes     = Union{Bool, Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32,
-                                 UInt64, UInt128, Float16, Float32, Float64}
+const SignedTypes        = [Int8, Int16, Int32, Int64, Int128]
+const UnsignedTypes      = [UInt8, UInt16, UInt32, UInt64, UInt128]
+const FloatTypes         = [Float16, Float32, Float64]
+const PrimitiveTypes     = [SignedTypes; UnsignedTypes; FloatTypes; Bool]
+const PrimitiveTypeTypes = Union{(Type{T} for T in PrimitiveTypes)...}
```
JonasIsensee, line 10: Is there a specific reason for this change?
It's not a problem, I'm just curious. I've never touched this code myself.

JonasIsensee:
Hi @eschnett,
thanks a lot for figuring this out!

src/data/number_types.jl
eschnett: This simplifies the code below. Before, it created a `Union` (instead of an array), and then use an internal method of `Base` to iterate over the types in that `Union`, and accessed the size of these types. The code is now simpler and doesn't use any undocumented functions any more. (One of these undocumented function was accessing the size of a type in a way that is not supported in Julia 1.9 any more.)

JonasIsensee: thank you

IanButterworth:
This issue resurfaced because julia compat was not updated https://github.com/JuliaRegistries/General/pull/93505

giordano:
It'd be appreciated if package developers proactively added appropriate compat bounds in the registry to avoid troubles to users, when they fix errors like this one.

JonasIsensee:
Hi @giordano ,


I'm aware of the problem but don't have a proper solution. 


Can I limit the compat on Julia (by default on the current minor version) without preventing tests on nightly?


Should I just manually keep track of problems and make handcrafted pull requests to the registry?

giordano:
We already fixed it in the PR linked above: https://github.com/JuliaRegistries/General/pull/93505.  That was more a suggestion for the future.

JonasIsensee:
Sure, I was just wondering whether I can prevent his from happening again or whether the best/only option is for me to create such a PR next time.

giordano:
Ideally you wouldn't use julia's internals. If that's really not possible, when you become aware of some breakage, like #428, it'd be good to do a PR to General like the one linked above, to prevent the package to be installed on incompatible versions of julia.

PR #446, opened 2023-01-14
allow loading compressed files during precompilation

marius311:
If you try to load a compressed file during precompilation (like top-level of your module, or in a `SnoopPrecompile.@precompile_all_calls`), the `@eval` s were causing an error since you're evaling into another module which is not allowed. You can just `getproperty` directly I think, works for my case, will wait for full suite of tests below to see if there's not some edge case.


marius311:
Any chance to get this merged/released? Anecdotally its been fine for me in all my work this week.

JonasIsensee:
Hi @marius311 ,
thanks for the ping. I just forgot about this. Will be tagging a release now.

Issue #454, opened 2023-02-16
Unpack for JLD2File


lukasgrunwald:
I would be nice to be able to use `@pack!` and `@unpack` from the `UnPack.jl`-package to quickly store and load values from JLD2File instances. This can be achieved by adding the methods
```
using UnPack
@inline UnPack.unpack(x::JLDFile{MmapIO}, ::Val{f}) where {f} =  x[string(f)]

@inline function UnPack.pack!(file::JLDFile{MmapIO}, ::Val{f}, val) where {f}
    file[string(f)] = val
end
``` 
After defining above extensions, these methods can be used as
```
using JLD2

x = rand(10)
path = "temp.jld2"
file = jldopen(path, "w")
@pack! file = x # Creates entry with name x
close(file)

file = jldopen("temp.jld2") # contains fields :x
@unpack x = file # Extracts x from the struct
close(file)
```

Maybe this could be added with UnPack.jl as on optional dependency; not sure how that works though? Happy to try and make a PR if you are open to doing this.

JonasIsensee:
Hi @lukasgrunwald ,
this is a neat idea.

I would recommend making the type signatures flexible enough to also accept `JLD2.Group`s.
You're welcome to give it a try.

Building a small extension like that isn't too hard. I'm not sure how much documentation there exists around it but
here's a link to some code where I tested out extensions for a different project https://gitlab.gwdg.de/eDLS/InPartS.jl/-/commit/a8982109c774ebc54933aaa95e4f9dcda1cdd1d4

PR #455, opened 2023-02-19
Add support for @unpack, @pack! for file-interface (+Test,Documentation)

lukasgrunwald:
Add optional support for @unpack, @pack! from [UnPack.jl](https://github.com/mauro3/UnPack.jl) for file-interface (see #454). They work for the normal file-interface and for `JLD2.Group`'s. The feature can be used as

```julia
using JLD2, UnPack
file = jldopen("example.jld2", "w")
x, y = rand(2)

@pack! file = x, y # equivalent to file["x"] = x; file["y"] = y
@unpack x, y = file # equivalent to x = file["x"]; y = file["y"]
```

The new methods are only added if the user loads the UnPack package himself. This is implemented via [Requires.jl](https://github.com/JuliaPackaging/Requires.jl). For Julia $\geq 1.9$ there are [Package extensions](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)), but since 1.9 is still in beta, I chose the Requires.jl version. The setup (with the extension living in `ext/UnPackExt.jl`) is compatible with the new "Extension-standard" and requires minimal updates, once this feature is stable. (I had some issues getting the Julia 1.9-beta to work with my VsCode, which is why I did it this way...)

In summary the PR implements:
- Add @(un)pack functionality, iff UnPack package is loaded (using Requires.jl)
- Add basic unit test for functionality
- Add documentation of feature
- Minor version bump (since new dependence)


lukasgrunwald:
Bump (the failed nightly runs are not due to any changes made in this commit)

README.md
```diff
@@ -99,6 +99,17 @@ jldopen("example.jld2", "w") do file
 end
 ```
 
+When additionally loading the [UnPack.jl](https://github.com/mauro3/UnPack.jl) package, the there defined `@unpack` and `@pack!` macros can be used to quickly save and load date from  the file-like interface. Example:
```
JonasIsensee, line 102:
```suggestion
When additionally loading the [UnPack.jl](https://github.com/mauro3/UnPack.jl) package, its `@unpack` and `@pack!` macros can be used to quickly save and load data from the file-like interface. Example:
```

README.md
```diff
@@ -142,6 +153,8 @@ or using slashes as path delimiters:
 @assert load("example.jld2", "mygroup/mystuff") == 42
 ```
 
+The group `file["mygroup"]` can be accessed with the same file-like interface as the "full" struct. When using [UnPack.jl](https://github.com/mauro3/UnPack.jl), the macros `@unpack` and `@pack!` work without modifications.
+
```
JonasIsensee, line 157: I think I'd prefer to have a separate section in the docs / readme titled something like `UnPack.jl API` 
rather than having it interleaved with the rest.

Personally, I haven't used these macros even once since the regular destructuring syntax was introduced.
(which can't really use here since it uses `getproperty`)

JonasIsensee (commented):
Hi @lukasgrunwald ,

thank you for contribution!

README.md
lukasgrunwald: Alright, good idea. Your suggestion is implemented in the new commit :+1:

README.md
lukasgrunwald: implemented in the new commit

JonasIsensee (approved):
Looks good to me! :)

PR #483, opened 2023-09-20
Don't keep file in open files cache if it fails to read metadata

Keno:
Fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1292


JonasIsensee:
Thanks!

I fixed the typo in the tests and tagged a patch release.

PR #492, opened 2023-10-17
Align writeas for Dict with wconvert

vtjnash:
Handles all cases, such as IdDict{K,V}. Seems to be likely the cause of the failure in https://github.com/JuliaLang/julia/pull/51319 / https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/abeb0f4_vs_3250804/JLD2.primary.log since it is trying to serialize the structure of the IdDict instead of the contents.

### DIFF
--- a/src/data/specialcased_types.jl
+++ b/src/data/specialcased_types.jl
@@ -246,9 +246,7 @@ rconvert(::Type{Core.SimpleVector}, x::Vector{Any}) = Core.svec(x...)
 
 ## Dicts
 
-writeas(::Type{Dict{K,V}}) where {K,V} = Vector{Pair{K,V}}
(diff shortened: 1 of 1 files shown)


sjkelly:
I'm not sure why CI is failing, but it is because this change.
Defining these functions in the REPL works.

oscardssmith:
This should have a test.

JonasIsensee:
This change looks fine. 
I find it rather curious that, so far, only IdDict{Any,Any} was covered...

JonasIsensee:
This breaks (de-) serialization of `<: AbstractDict` that have additional fields.
#497

The custom serialization should not be defined for all `<: AbstractDict` even if it seems convenient at first.

PR #493, opened 2023-10-17
switch to PrecompileTools.jl

sjkelly:
This provides a minor TTFX speedup, for example with the workload
in src/precompile.jl:

```
  julia> @time mktemp() do path, _

               # jldsave
               jldsave(path; a, b, c, d, e, f, g, h)

               jldopen(path) do f
                   for k in keys(f)
                       f[k]
                   end
               end

               nothing
           end
```

Before:

```
  0.158051 seconds (195.66 k allocations: 13.735 MiB, 33.56% gc time, 95.68% compilation time)
```

After:

```
  0.038961 seconds (24.42 k allocations: 1.921 MiB, 80.58% compilation time)
```

The load time/memory is essentially the same:

Before:

```
  julia> @time using JLD2
  0.102329 seconds (137.15 k allocations: 12.254 MiB)
```

After:

```
  julia> @time using JLD2
  0.106296 seconds (146.97 k allocations: 12.871 MiB)
```


JonasIsensee (approved):
I guess, I haven't looked closely enough at precompiletools.jl to understand why this helps but since it does and tests pass.

Thanks !

PR #509, opened 2023-12-11
compat: no MmapIO for windows 7

HongBinYu-hub:
Ensure win7 can be precompiled
Win7 and below systems use IOStream method to ensure pre-compilation correct.

### DIFF
--- a/src/JLD2.jl
+++ b/src/JLD2.jl
@@ -56,6 +56,14 @@ include("bufferedio.jl")
 include("misc.jl")
 include("superblock.jl")
 
+is_win7() = Sys.iswindows() && Sys.windows_version().major <= 6 && Sys.windows_version().minor <= 1
+# Windows 7 doesn't support mmap, falls back to IOStream
+const DEFAULT_IOTYPE = if is_win7()
+    IOStream
+else
+    JLD2.MmapIO
+end
+
 """
     RelOffset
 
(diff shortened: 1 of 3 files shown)


johnnychen94:
I really think you should change the default value for `iotype` (everywhere) to be `DEFAULT_IO_TYPE`, where 

```Julia
# Windows 7 doesn't support mmap, falls back to IOStream
const DEFAULT_IO_TYPE = if is_win7()
    IOStream
else
    MmapIO
end

is_win7() = Sys.iswindows() && Sys.windows_version().major <= 6 && Sys.windows_version().minor <= 1
```

johnnychen94:
@JonasIsensee Are you okay with this compatibility patch? For some reason, we ([Suzhou Tongyuan](https://github.com/Suzhou-Tongyuan)) need to keep compatibility with some "strange" platforms, e.g., windows 7.
We prefer to contribute upstream first, so here's the patch.

If you don't want to get this patch in, we understand, and we'll instead start to maintain our internal fork version.

src/JLD2.jl
```diff
@@ -56,6 +56,14 @@ include("bufferedio.jl")
 include("misc.jl")
 include("superblock.jl")
 
+is_win7() = Sys.iswindows() && Sys.windows_version().major <= 6 && Sys.windows_version().minor <= 1
+# Windows 7 doesn't support mmap, falls back to IOStream
+const DEFAULT_IOTYPE = if is_win7()
+    IOStream
+else
+    JLD2.MmapIO
```
johnnychen94, line 64: I think it should be

```suggestion
    MmapIO
```

johnnychen94 (approved):

JonasIsensee:
> @JonasIsensee Are you okay with this compatibility patch? For some reason, we ([Suzhou Tongyuan](https://github.com/Suzhou-Tongyuan)) need to keep compatibility with some "strange" platforms, e.g., windows 7. We prefer to contribute upstream first, so here's the patch.
> 
> If you don't want to get this patch in, we understand, and we'll instead start to maintain our internal fork version.

I apologize for the late reply..
I like this approach:
`const DEFAULT_IO_TYPE = if is_win7()`

johnnychen94:
Thanks for merging!

I'll use my privilege to make a patch release #514

PR #517, opened 2023-12-19
correct canbeuninitialized

JonasIsensee:


JonasIsensee:
This shows a a new problem on nightly but this probably already existed beforehand and just wasn't visible due to previous errors.

PR #569, opened 2024-07-15
Don't hide exception data during loading and saving

Octogonapus:
This PR removes some rethrows which cause the real exception to be hidden (which makes using JLD2 more difficult
because I don't know the real reason an exception happened).


JonasIsensee:
Thanks!
Not sure what I was thinking when writing those lines...

PR #582, opened 2024-07-30
WIP: Mmappable Arrays

### LINKED ISSUE #25: Align bitstype arrays in file
This would be nice if we ever want to support `mmaparrays` here, which is currently accepted as an option but ignored.

### LINKED ISSUE #71: Option to read slices of stored arrays
Is it possible to read parts of arrays stored as JLD2 files (a specific slice), like in the HDF5.jl example?
```
using HDF5
h5read("/tmp/test2.h5", "mygroup2/A", (2:3:15, 3:5))
```

### LINKED ISSUE #124: change item in existing collection
Hi, 
I am using JLD2 to save the results of scientific computations and I wonder if there is a way to modify an array inside a JLD file. 
Here is an example of what I would like to do
~~~
using JLD2
myarray= [1 2 3]
@save "testfile.jld2" myarray
jldopen("testfile.jld2", "r+") do myfile
    myfile["myarray"][2]=4
end
jldopen("testfile.jld2", "r") do myfile
    myfile["myarray"]
end
~~~
and I would like to get [1 4 3] instead of the current [1 2 3]. 
I found that it is possible to make the change persist by opening the file with h5open but not with jldopen. 
Is there a way to make such persistent changes in JLD2 ?

JonasIsensee:
Aims to make mmap for reading and **editing** (!) Arrays in files possible.

- Requires 8-byte aligned address -> this was not ensured previously. Arrays with current JLD2 will have a 1/8 chance of being mmappable
- Requires `JLD2.samelayout(type) == true`

Currently adds API:
- `ismmappable`: Test mmappability
- `mmaparray`: Load array dataset as mmapped.
- `allocate_early(dset::Dataset, ::Datatype)`: Write array dataset to file without providing the actual array yet. Just reserve space. Can be opened using `mmaparray` and filled with data afterwards.

closes #124
closes #71 (by taking slices of a mmapped array)
closes #25


cstjean:
Two questions:

- Does this only work for `isbits` types?
- I thought that HDF5 wasn't good at deleting / replacing data. I guess you're getting around that since this is a really simple overwriting (assuming it's `isbits`?)

JonasIsensee:
> Two questions:
> 
>     * Does this only work for `isbits` types?
> 
>     * I thought that HDF5 wasn't good at deleting / replacing data. I guess you're getting around that since this is a really simple overwriting? (assuming it's `isbits`)

Yeah, so the restriction is that this only works when `JLD2.samelayout(type) == true`. This actually even more restrictive than `isbits`. The memory layout needs to be identical. This is only true if the type has a size that is a multiple of 8. (Otherwise julia will pad the type at the end to become a multiple of 8).

Yes, deleting *efficiently* is super hard, because you end up with little bits of empty space in different place in the file and would need to keep track of it in order to reuse it.
Here, this is not a problem, because mmap allows in-place editing of the array that is a fixed-size contiguous chunk of bytes in the file.

PR #591, opened 2024-09-04
v0.5.0

JonasIsensee:


ChrisRackauckas:
What is breaking in this change?

JonasIsensee:
This is breaking at the file format level.
New files cannot be read by older versions of the library.
Existing files can still be read and there are no breaking API changes.

ChrisRackauckas:
Can you add that to the news?

JonasIsensee:
It is in the changelog.
I can copy it into the description of this PR and into the release notes ,I guess

PR #613, opened 2024-11-04
Add warning, as and copied from BSON.jl.

PallHaraldsson:


PallHaraldsson:
It seems good to let people know there (as soon as possible), but not at target:

https://github.com/JonasIsensee/JLD2DebugTools.jl  is archived, so further changes may be needed to the docs, though I think it still works. Could be done in different (or same PR, if people want to suggest adding to it here soon).

JonasIsensee:
Hi!

Thank you for this suggestion.

Aside from people providing `.jld2` files to me for debugging purposes I have not seen JLD2 files being distributed on the internet. That is part of the reason that there is no large warning at the top of the readme.
The other part is that this is a much more generic problem that exists for the serialization stdlib as well. (and any package installation...)

I suppose this also addresses your question on the default: The default is the "non-safe" version because otherwise (I would guess) 99% of users / use cases would need to explicitly add `plain=false` to their load calls.

PR #615, opened 2024-11-06
Nightlyfix

### LINKED ISSUE #614: Incompatibility with Julia 1.12 due to missing `Base.module_keys`
There are a couple places where JLD2 currently depends on `Base.module_keys`:
https://github.com/search?q=repo%3AJuliaIO%2FJLD2.jl%20module_keys&type=code

This function is no longer available on nightly (1.12) versions of Julia, which makes JLD2 incompatible.

JonasIsensee:
fixes #614

### DIFF
--- a/src/data/reconstructing_datatypes.jl
+++ b/src/data/reconstructing_datatypes.jl
@@ -308,9 +308,7 @@ function _resolve_type(rr::MappedRepr{T,DataTypeODR},
                        hasparams::Bool,
                        params) where T
     parts = split(mypath, '.')
-    modules = vcat([Main], collect(keys(Base.module_keys)))
-    unique!(modules)
-    for mod in modules
+    for mod in values(Base.loaded_modules)
         resolution_attempt = _resolve_type_singlemodule(rr,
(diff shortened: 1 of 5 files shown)


src/data/reconstructing_datatypes.jl
```diff
@@ -308,9 +308,7 @@ function _resolve_type(rr::MappedRepr{T,DataTypeODR},
                        hasparams::Bool,
                        params) where T
     parts = split(mypath, '.')
-    modules = vcat([Main], collect(keys(Base.module_keys)))
-    unique!(modules)
-    for mod in modules
+    for mod in values(Base.loaded_modules)
```
vtjnash, line 311: You should preferably use `Base.loaded_modules_array`, as that has been the stable and safe API for this for many versions (since JuliaLang/julia#23579), while this is neither safe nor stable

JonasIsensee: Nice! I didn't know that function existed.
Thanks, I'll switch to that in the next release.

PR #626, opened 2024-12-30
Cleanup of Lookup3 hash

nhz2:
This PR uses the `bitrotate` function added in Julia v1.5, it also changes a `convert(UInt32, n)` to `n%UInt32` when setting the initial state to match what happens in the C code. Lastly, it adds a missing `GC.@preserve` around where a pointer to an array is used.


JonasIsensee:
Thanks for this cleanup!

PR #631, opened 2025-02-10
Backports some fixes to 0.4

KristofferC:
A large number of packages seems to be stuck on JLD2 0.4 and they error on upcoming 1.12 with 

```
[0KERROR: LoadError: UndefVarError: `module_keys` not defined in `Base`
[0KStacktrace:
[0K  [1] getproperty
[0K    @ ./Base_compiler.jl:47 [inlined]
```

so I think it is a good idea to fix things on 0.4 as well. To make the tests pass I also had to backport the loading decompressor deprecation.


JonasIsensee:
Thank you, this is a reasonable first step.
I'll also see if I can find which packages are being held back and what keeps them from updating compat bounds.

PR #635, opened 2025-03-20
convert FileIO and UnPack to true extensions

vtjnash:
Avoids the recursion of having FileIO expect to load JLD2 in the precompile script, triggering possible errors. The functions load and save were not previously documented as part of the API (although they still exist here as public functions when desired), and the tests even usually explicitly got them from FileIO when desired.

This is intended to fix CI for v1.12+ (dropping support for v1.8 and earlier which are EOL anyways since v1.9 add extensions)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,19 +3,26 @@ uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 version = "0.5.11"
 
 [deps]
-FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 TranscodingStreams = "3bb67fe8-82b1-5028-8e26-92a6c54297fa"
 
+[weakdeps]
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
+
+[extensions]
+FileIOExt = "FileIO"
+UnPackExt = "UnPack"
+
 [compat]
-FileIO = "1.5"
+FileIO = "1"
 MacroTools = "0.5.10"
+Mmap = "1"
 OrderedCollections = "1"
 PrecompileTools = "1"
-Requires = "1.3"
 TranscodingStreams = "0.9, 0.10, 0.11"
-julia = "1.6"
+UnPack = "1"
+julia = "1.9"
--- a/ext/FileIOExt.jl
+++ b/ext/FileIOExt.jl
@@ -1,0 +1,10 @@
+module FileIOExt
+
+import JLD2: fileio_load, fileio_save
+using JLD2: load, save
+using FileIO: FileIO, File, @format_str, filename
+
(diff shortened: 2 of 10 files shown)


JonasIsensee:
So far, there were no good reasons to drop support for v1.6. Now is the time, it seems.

Hm, i was sure, save and load appeared in the docs somewhere. 
Anyway, thanks for these changes!

`public` probably needs to be put in an appropriate version `if` statement, though.

JonasIsensee:
Hmm, `load` and `save` used to be explicitly re-exported.
That makes this a breaking change.

vtjnash:
While in the `export` list, they are not documented as being there, so I'd argue this is actually conforming the implementation to the documentation rather than breaking

JonasIsensee:
image

Project.toml
```diff
@@ -3,19 +3,26 @@ uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
...
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
+
+[extensions]
+FileIOExt = "FileIO"
+UnPackExt = "UnPack"
+
 [compat]
-FileIO = "1.5"
+FileIO = "1"
```
JonasIsensee, line 21:
```suggestion
FileIO = "1.5"
```

JonasIsensee:
Do I understand correctly that the real problem here is the package loading shenanigans that does - which I currently try to use as part of the precompile script?

Would it then not be better to just remove this particular precompile step?
(and and directly precompile `JLD2.fileio_load`)

vtjnash:
Alright, I see https://github.com/JuliaIO/JLD2.jl/pull/300 that it is intentional recently this package added a non-weak dependency on FileIO and added load/save to the exports again (in contrast to older PRs that had argued it should not be exported). I can adjust this to make FileIO a strong dependency again if that is preferred

JonasIsensee:
If this problem for v1.12 can be fixed with a modification to the precompile file, then that is what I would prefer over a change of semi-publicly declared API.

I've opened #637 to test this

JonasIsensee:
> Alright, I see #300 that it is intentional recently this package added a non-weak dependency on FileIO and added load/save to the exports again (in contrast to older PRs that had argued it should not be exported). I can adjust this to make FileIO a strong dependency again if that is preferred

Indeed, I believe I changed my mind at least once over the past years on whether I liked the FileIO API or not.
As of #300, my position was that a unified API is worth the effort, though I am unsure whether the other IO packages have since moved on from that ideal.

[ I would like to apologize, if I am currently coming across as a bit impatient / cranky / uncooperative -- I will be submitting my PhD thesis in 5 weeks and my mind is a preoccupied with that currently ]

vtjnash:
The change in https://github.com/JuliaIO/JLD2.jl/pull/637 sounded good to me. Best of luck on your thesis, and thanks for taking the time to review and discuss. I'll let you decide if you want to do anything else with the suggested changes here later. I think the only one of any relevance is replacing Requires.jl with an extension, which is mostly a mechanical change here.

t-bltg:
Can this PR get in since [julia 1.12.0-beta1](https://discourse.julialang.org/t/julia-v1-12-0-beta1-is-now-available/127686) is out ?

Failing CI for Plots : https://github.com/JuliaPlots/Plots.jl/actions/runs/14552071254/job/40824448850#step:9:2002.

JonasIsensee:
> Can this PR get in since [julia 1.12.0-beta1](https://discourse.julialang.org/t/julia-v1-12-0-beta1-is-now-available/127686) is out ?
> 
> Xref failing CI for Plots : https://github.com/JuliaPlots/Plots.jl/actions/runs/14552071254/job/40824448850#step:9:2002.

This PR does a few more things which I am not entirely happy about.
I just made #643 which only removes requires and creates an Extension for UnPack.jl.
I will release that as soon as tests pass and that should also fix your CI failure, I hope.

t-bltg:
Awesome @JonasIsensee, thanks for taking this issue forward :+1:

PR #637, opened 2025-03-21
stop precompiling FileIO.load

JonasIsensee:


KristofferC:
New version with this would be nice

JonasIsensee:
> New version with this would be nice

should be out any time now.

PR #650, opened 2025-06-24
Advanced typemap and full user control

### LINKED ISSUE #630: Typemap yields `Upgrade(type)` objects on load
Thank you for answering #627. So far it works well, however I found an issue:

```julia
using JLD2

struct Foo
    x::Int
end

jldsave("test.jld2"; type=Foo)

load("test.jld2", "type"; typemap=Dict("Main.Foo"=>JLD2.Upgrade(Foo)))
```
This should reasonably yield `Foo`, but it yields `Upgrade(Foo)`. To put it another way, if I'm deserializing a `Dict(Foo=>4)`, it'll become `Dict(Upgrade(Foo)=>4)`.

JonasIsensee:
This PR gives the user full control over type reconstruction.

Simply pass a function `(f::JLDFile, typepath::String, params::Vector) -> type` as keyword to `load` or `jldopen`.
This function is then called for every single type that will be decoded.
The default implementation is `JLD2.default_typemap` and all standard cases can simply be forwarded to it.

Example: 
Here, for bogus reasons we want to load a `Complex{Int}` as a `Complex{Float64}` while also modifying the real part during loading.
`if typepath == "Base.Complex" && params == [Int]` checks that only `Complex{Int}` get changed and 
wrapping the updated type in `JLD2.Upgrade` ensures that our custom implementation of `JLD2.rconvert` gets called during loading.

```
    fn = "advancedtypemap.jld2"
    z = 1 + 2im # Complex{Int}
    zf = 1f0 + 2f0*im # Complex{Float32}
    jldsave(fn; z, type = Complex{Int}, zf)

    typemap = function(f, typepath, params)
        @info "typepath: $typepath, params: $params"
        if typepath == "Base.Complex" && params == [Int]
            return JLD2.Upgrade(Complex{Float64})
        end
        JLD2.default_typemap(f, typepath, params)
    end

    JLD2.rconvert(::Type{Complex{Float64}}, nt::NamedTuple) =
        Complex{Float64}(nt.re + 1, nt.im)

    d = load(fn; typemap)
    @test d["z"] isa Complex{Float64}
    @test d["z"] == 2.0 + 2.0im
    @test d["type"] == Complex{Float64}
    @test d["zf"] isa Complex{Float32}
    @test d["zf"] == 1.0 + 2.0im
```

closes #630

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.5.15
+ - Add full control over type mapping by providing a typemap function
+
 ## 0.5.14
  - Add support for `Memory` and `MemoryRef`
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JLD2"
 uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
-version = "0.5.14"
+version = "0.5.15"
 
 [deps]
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
--- a/docs/src/advanced.md
+++ b/docs/src/advanced.md
@@ -3,7 +3,7 @@
 ## Explicit Type Remapping
 
 Sometimes you store data using `struct`s that you defined yourself or are
-shipped with some package and weeks later, when you want to 
+shipped with some package and weeks later, when you want to
 load the data, the structs have changed.
 
 ```julia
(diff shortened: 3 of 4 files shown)


cstjean:
Thanks! I'm assuming this is called only once per type + parameters, and cached? Otherwise, the type instability looks brutal to have that applied to every single object.

PR #651, opened 2025-06-27
compression rework

JonasIsensee:
This PR  aims to completely overhaul how compression filters work in JLD2 and tries to imitate the approach of HDF5.jl.

Dynamic loading of compression codecs is removed as it has stopped working reliably on recent julia versions.

On top of that, JLD2 no longer asks the user to do e.g. `CodecZlib` but instead introduces small packages for each filter such as `JLD2Deflate`.
These export the filters (e.g. `Deflate()`) are mostly empty and the real implementations sit within package extensions of JLD2 to prevent potential version conflicts.

Preliminary API could look like:

```
using JLD2, JLD2Deflate
arr = rand(1:10, 10000)
save("test.jld2", "arr", arr; compress=[Shuffle(), Deflate()])
```

Tests can't work because I haven't figured out how to test without the wrapper pkgs being registered.

This is also still missing lots of docs and attribution. 
In particular for the not-yet added filters (blosc, bitshuffle, lz4, bzip2)), I will try to copy as much as possible from the HDF5.jl implementations.

Now has implementations for 
`Deflate` (Zlib)
`Shuffle`
`Zstd`
`Lz4` 
`Bzip2`
`Bitshuffle`
`Blosc`

Note: The current implementation is a little more happy to throw errors compared to the hdf5 version. 
HDF5 can mark filters as "inactive" for indivdual chunks of a larger (chunked) array, which is useful e.g. when a compression filter is ineffective.
Writing chunked arrays isn't implemented in JLD2 so far, so this does not matter too much just yet but might be good to add later.
(HDF5 does this via the return value of the of the compression function)

#### How to test:
Checkout this PR and also `dev` the filter pkgs with e.g.
```
using JLD2, Pkg
Pkg.develop(joinpath(pkgdir(JLD2),  "filterpkgs", "JLD2Deflate"))
Pkg.develop(joinpath(pkgdir(JLD2),  "filterpkgs", "JLD2Blosc"))
using JLD2Deflate, JLD2Blosc
``` 
The docs build that belongs to this PR  already describes how to use the new API: https://juliaio.github.io/JLD2.jl/previews/PR651/

I'm interested in your opinions and feedback  on this interface.

#### Update:
`Zlib_jll` will be shipped as a stdlib starting from julia 1.12. 
Therefore, I have decided to make it a hard dependency rather than a weak one.
Thus the default `Deflate()` compression will not require loading an additional filter pkg.


JonasIsensee:
@mkitti This PR obviously takes heavy inspiration from HDF5.jl in design but the filter implementations are also adaptations of your code.
I tried to make this clear using e.g. the thirdparty.md files and in comments.
Beyond overall feedback, I'd be particularly interested in hearing your opinion if this attribution feels sufficient to you.

nhz2:
For memory safety reasons, I think blosc and bitshuffle support should be delayed until some upstream bugs are fixed. Ref: https://github.com/Blosc/c-blosc/issues/385 https://github.com/JuliaIO/ChunkCodecs.jl/issues/58#issuecomment-3045435721

mkitti:
We should strongly consider not encoding new data with c-blosc (v1) and rather encode new data with c-blosc2. c-blosc2 is seeing active development where as c-blosc development is mostly responding to community bug reports.

nhz2:
Another API to compare with is how filters are registered in go's zip standard library https://github.com/golang/go/blob/master/src/archive/zip/register.go

Ideas we can copy here are:
1. Registering and looking up the compressor/decompressor is thread-safe.
They are using a [`sync.Map`](https://github.com/golang/go/blob/fca43a8436f2dbec981fb002c6ba31fe9ff89dd6/src/archive/zip/register.go#L104-L107), not sure what the best Julia equivalent would be.
2. Different types and registries for compression and decompression. It may make sense to split these here as well.

JonasIsensee:
> We should strongly consider not encoding new data with c-blosc (v1) and rather encode new data with c-blosc2. c-blosc2 is seeing active development where as c-blosc development is mostly responding to community bug reports.

Is there hdf5 precedent for using c-blosc2 ?
Would this use the same filter id?
(The c-blosc2 repo says, that it is backward but not forward conpatible...)

JonasIsensee:
> Yes, the error handling for EncodingOptions is in the constructor; after that, the actual encoding should ideally never error if the input data has a valid size.

Alright, that's a plausible but incompatible design decision.

That means, it's either something like
`Filter(zstd, (;compressionLevel= 4, checksum= true))`

or,  JLD2 just ships it's own API using ChunkCodec under the hood.

nhz2:
I think it is worth considering a completely extension free way of doing things.
If the goal is to avoid the `JLD2Lz4` wrapper packages, the functions and abstract types defined in `JLD2` and specialized in `JLD2Lz4` could be moved into functions and abstract types defined in `ChunkCodecCore` and specialized in `ChunkCodecLibLz4`.

Also if decoding filter registration is explicit, there is also no need for the `__init__` stuff in the extensions. You will just get an error like:

```
Error: decoder for lz4, id 32004, is not available. Install ChunkCodecLibLz4.jl and run the following to register a decoder:
using ChunkCodecLibLz4: LZ4_JLD2
JLD2.register_decoder!(32004, LZ4_JLD2)
```

Also XRef: https://github.com/SciML/DifferentialEquations.jl/issues/1082#issuecomment-2773506976 which has a good discussion of using extensions for optional dependencies.

JonasIsensee:
> Also XRef: [SciML/DifferentialEquations.jl#1082 (comment)](https://github.com/SciML/DifferentialEquations.jl/issues/1082#issuecomment-2773506976) which has a good discussion of using extensions for optional dependencies.

a good but worrying discussion.

> Because JLD2 is a dependency of DrWatson.jl, it is likely to be loaded in the global environment, while the extension trigger is loaded in a later activated environment.

This is just as troublesome.

---

Ok, let me summarize my understanding so far:

 * `JLD2` will often be loaded as part of stacked environments (DrWatson.@quickactivate and friends).
 *  These are known to cause uncontrolled loading of incompatible versions of packages.
 * These are also known to mess up extension loading.

* Broken extension loading is very intransparent from the user side and should be avoided.
* Risks of incompatible versions between pkgs can be minimized by having good interfaces that rarely break (lowering the chances of versions actually being incompatible..)
* One could potentially introduce a run-time check inside an `__init__()` function of e.g. `JLDLz4` to warn about an incompatible `JLD2` version. 

* I need *either* an extension or a filter pkg `JLDLz4` to be able to specify compat bounds (!) on the ChunkCodec implementation.
Simply depending on `ChunkCodecCore` and hoping that all `CCLibXXX` pkgs are compatible by transitive compat bounds seems fishy. Would only work reliably if `ChunkCodecCore` gets exact lower compat bounds and also gets a new release for every bugfix in each of the downstream `ChunkCodecLizXX` pkgs. 

* I am not against registering a few pkgs like `JLD2Lz4`. I just wanted to be sure it's a good thing to do.
Also, given that `Zlib` and `Zstd` can now be real deps / builtin filters as the binary dep will be stdlibs from 1.12 onward,
and that `Blosc` and `Bitshuffle` will be delayed, It is just `JLD2Lz4` and `JLD2Bzip2` for now.

* `HDF5.jl` went the opposite way: from external filter pkgs  to wrapper pkgs with extensions for the same reason of compat headaches. However, AFAICT the situation was different. `HDF5` likely does not encounter the stacked environment issue as often. (Not a dep of DrWatson)

---

My takeaway is to create pkgs `JLD2Lz4` and `JLD2Bzip2` that are not just wrappers but also contain the implementation - no extensions.
Try hard, to rarely break the interfaces, so version conflicts (from inconsistent environments) just don't happen.

JonasIsensee:
> > > * They are using a [`sync.Map`](https://github.com/golang/go/blob/fca43a8436f2dbec981fb002c6ba31fe9ff89dd6/src/archive/zip/register.go#L104-L107), not sure what the best Julia equivalent would be.
> > 
> > 
> > There's `ConcurrentDict` from ConcurrentCollections.jl. We could also put a lock around it via `Base.Lockable(Dict())` in Julia 1.11.
> 
> Any thoughts on pirating methods on `Val{id}` like: `JLD2.get_filter(::Val{32004}) = LZ4Filter` as an alternative to `__init__() = register_filter(Lz4Filter)` ?

That would work. I thought about doing that and it would allow us to get rid of the odd dictionary logic.

I'm wondering, if it is a good idea to not even maintain a single list of the available filters.
It might be nice to have a function like `JLD2.available_filters()` at some point.
This could, of course, just be `available_filters() = print(subtypes(Filter))`.

JonasIsensee:
> [JuliaIO/ChunkCodecs.jl#63](https://github.com/JuliaIO/ChunkCodecs.jl/pull/63) adds support for Julia 1.9.

Cool!

> Besides element type and chunk dimensions, I can't think of anything else a filter would use for now, but it's good to allow this to be expanded in the future. To make expanding this in the future easier, the second `Filters.compress` method should have a different name like `compress_array`, so plugin packages can check `isdefined(JLD2, :compress_array)` before specializing for backwards compatibility. Also, what is `odr::S` in your example?

I made this part of the internal api match the hdf5 one a bit more closely. This should work for now and can be extended later, since the two external pkgs, (Lz4 and Bzip2) don't use these features anyway.

> In terms of error handling, HDF5 allows filters in a pipeline to be marked as optional, so if a chunk is incompressible, that filter can be skipped for that chunk. For reading data, I think it would be fine to error if the chunk is undecodable, because that would indicate that the file is probably corrupted.

Writing data into multiple chunks is not yet supported, but I've extended the API to report return codes that can later be used to mark non-compressed chunks.

JonasIsensee:
> My 2 cents is that stacked environments are fundamentally dangerous. It's a shame if this makes life harder for package writers. Why not recommend adding DrWatson explicitly to the environment, and maybe rely on https://github.com/LilithHafner/BasicAutoloads.jl for convenience?

Adding DrWatson to the environment might help but mostly through "luck".
It does not change that `@quickactivate` promotes the idea of changing the environment from within a running session that has pkgs already loaded. (at DrWatson..).

I'd go one step further and say that `Pkg.activate` is fundamentally dangerous and should not be used for anything other than setting up environments for the first time and never to then load pkgs and run code.

PR #655, opened 2025-07-23
Fix hash function definition

bvdmitri:
As said in the [documentation](https://docs.julialang.org/en/v1/base/base/#Base.hash)

> New types should implement the 2-argument form, typically by calling the 2-argument hash method recursively in order to mix hashes of the contents with each other (and with h).

Otherwise, it causes [invalidations](https://julialang.org/blog/2020/08/invalidations/). Initially spot on in RxInfer (that uses GraphPPL that uses MetaGraphsNext that uses JLD2) when analyzing latency issues on upcoming Julia 1.12. Would be nice to tag a new patch release after this is merged 

Before this PR 
image

After this PR
image

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -148,7 +148,7 @@ define_packed(RelOffset)
 
 RelOffset(r::RelOffset) = r
 Base.:(==)(x::RelOffset, y::RelOffset) = x === y
-Base.hash(x::RelOffset) = hash(x.offset)
+Base.hash(x::RelOffset, u::UInt) = hash(x.offset, u)
 Base.:(+)(x::RelOffset, y::Integer) = RelOffset(UInt64(x.offset + y))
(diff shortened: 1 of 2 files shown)


src/types.jl
```diff
@@ -148,7 +148,7 @@ define_packed(RelOffset)
 
 RelOffset(r::RelOffset) = r
 Base.:(==)(x::RelOffset, y::RelOffset) = x === y
-Base.hash(x::RelOffset) = hash(x.offset)
+Base.hash(x::RelOffset, u::UInt) = hash(x.offset, u)
```
KristofferC, line 151: This is almost always called `h`

KristofferC (approved):

JonasIsensee:
Thanks to both of you!

PR #660, opened 2025-08-25
Fix JLD2Lz4 backwards compatibility

nhz2:
Ref https://github.com/JuliaIO/JLD2.jl/issues/659


nhz2:
Yes `JLD2TestFiles` is a better place for it.

PR #667, opened 2025-09-01
Compression doc fixes

nhz2:
This PR makes the compression doc examples runnable and adds some more details on filter compatibility.


docs/src/compression.md
```diff
@@ -145,29 +145,47 @@ settings.
 ```julia
 using JLD2
 
-jldopen("example.jld2", "w") do f
+jldopen("example.jld2", "w"; compress=ZstdFilter()) do f
```
JonasIsensee, line 148: It seems a bit counter-intuitive that you provide the compress argument to the whole file and then directly overrule it in the first few `write` calls.
Maybe it would be better to move the `default_array` to the top?

JonasIsensee (commented):
Thanks for reading the docs so carefully and providing improvements!

Issue #659, opened 2025-08-25
LZ4 compatibility


nhz2:
JLD2 0.6.0 is unable to read `LZ4FrameCompressor` compressed data from 0.5.15.

To test this, I generated the following file using the included script

[oldjld2.tar.gz](https://github.com/user-attachments/files/21961500/oldjld2.tar.gz)

We probably need https://github.com/HDFGroup/hdf5_plugins/issues/134#issuecomment-2484434118 to ensure the new version can still read the previous format.

I'm not sure how older versions of JLD2 will be able to read the new version.

mkitti:
What can we do in terms of detection of the format? Is there a heuristic that we can use to detect the on-disk format.

nhz2:
Yes, we can check for the b"\x04\x22\x4D\x18" lz4 frame magic to know it is the old format.

PR #668, opened 2025-09-01
JLD2Lz4: Add check for LZ4 Frame magic during encoding

nhz2:
This is more of a comment because this error cannot be triggered on any existing hardware.

Ref #659

### DIFF
--- a/filterpkgs/JLD2Lz4/src/JLD2Lz4.jl
+++ b/filterpkgs/JLD2Lz4/src/JLD2Lz4.jl
@@ -38,7 +38,7 @@ function Filters.apply_filter!(filter::Lz4Filter, ref::Ref, forward::Bool=true)
         # This is never going to error unless someone can encode about 
         # 300 PB in a single chunk.
         if UInt64(length(buf))>>32 == 0x0422_4D18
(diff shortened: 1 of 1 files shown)


filterpkgs/JLD2Lz4/src/JLD2Lz4.jl
```diff
@@ -32,6 +32,14 @@ Filters.filtertype(::Val{32004}) = Lz4Filter
 
 function Filters.apply_filter!(filter::Lz4Filter, ref::Ref, forward::Bool=true)
     if forward
+        buf = ref[]
+        # This check is to avoid confusion with LZ4 Frame magic
+        # for backwards compatibility with pre 0.6 JLD2 files
+        # This is never going to error unless someone can encode about 
+        # 300 PB in a single chunk.
+        if UInt64(length(buf))>>32 == 0x0422_4D18
+            error("Encoding would be confused for LZ4Frame format.")
```
JonasIsensee, line 41: How about this?

```suggestion
            throw(UnsupportedFeatureException("This particular dataset size is disallowed for the Lz4Filter. Please use a different compression filter."))
```

PR #669, opened 2025-09-01
Avoid mutating filters.

nhz2:
This PR changes the mutating function `set_local!(filter, odr, dataspace, datasetcreationprops)` to the non-mutating
`set_local(filter, odr, dataspace, datasetcreationprops)` and also makes `Shuffle` a `struct` instead of a `mutable struct`.

This is not a breaking change because `set_local!` is not used by any plugins yet, and currently, the internal filter interface is still experimental.

I imagine people will create global variables with filter options like:

```
const filters = [Shuffle(), Deflate()]
```

And then reuse this in multiple threads. With the current `set_local!` function for `Shuffle`, this would be a race condition.
This PR fixes that issue.

### DIFF
--- a/src/Filters.jl
+++ b/src/Filters.jl
@@ -61,11 +61,11 @@ function apply_filter! end
 
 
 """
-    set_local!(filter, odr, dataspace, datasetcreationprops)
+    set_local(filter, odr, dataspace, datasetcreationprops) -> filter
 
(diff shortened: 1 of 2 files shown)


JonasIsensee:
Interesting.
I didn't think anyone would try to do something like that, but you are right. It's possible and your suggested version is a bit cleaner anyway. 👍

PR #672, opened 2025-09-05
Improve filter docs

nhz2:
This PR adds a docstring for `Bzip2Filter`, fixes some urls, and also shortens the `filtername` for the filters.

The `filtername` is for debug purposes.
Since the `filtername` is stored in the .jld2 file, a long name adds to the overhead of using compression on small arrays.

### DIFF
--- a/src/Filters.jl
+++ b/src/Filters.jl
@@ -182,6 +182,7 @@ struct Shuffle <: Filter
     element_size::UInt32
 end
 Shuffle() = Shuffle(0)
+filtername(::Type{Shuffle}) = ""
 filterid(::Type{Shuffle}) = UInt16(2)
 client_values(filter::Shuffle) = (filter.element_size,)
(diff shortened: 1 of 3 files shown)


src/Filters.jl
```diff
@@ -182,6 +182,7 @@ struct Shuffle <: Filter
     element_size::UInt32
 end
 Shuffle() = Shuffle(0)
+filtername(::Type{Shuffle}) = ""
```
JonasIsensee, line 185: This definition would not be used.
Filters with an id `<=255` do not have a filtername by definition of the h5 format spec.

JonasIsensee (approved):
Thanks for the improvements!
lengthy filter names were directly copied over from HDF5.jl but I don't think they are actuallly ever used for anything meaningful, so I'm fine with shortening them.

PR #705, opened 2025-12-20
`Bzip2Filter` constructor doesn't support keyword arg, update docs accordingly

BioTurboNick:


JonasIsensee:
I finally got around to actually releasing the fix.

PR #718, opened 2026-06-04
compat: Support the TypeEq type-system refactor (JuliaLang/julia#61915)

Keno:
The TypeEq refactor changes `typeof(Type{T})` from `DataType` to the new `Core.TypeEq`. As a consequence, a `Type{T}` value is no longer an instance of `DataType` (`isa(Type{T}, DataType)` is now `false`), even though `Type{T} <: DataType` still holds and `Base.isType(Type{T})` is `true`.

JLD2 dispatches its `DataType` serialization machinery on `typeof(x)`, so on such a Julia a value whose type is a `Type{T}` (for example the field of a struct declared `x::Type{Float64}`) was routed through the generic compound path instead of the dedicated `DataType` path. This surfaced as a `MethodError` in `commit` and downstream `fieldcount`/`convert` errors when writing such values.

Introduce a `typeof_serializable` helper that maps `Core.TypeEq`-typed values back to `DataType` (guarded by `isdefined(Core, :TypeEq)` so the code still loads on released Julia), and relax the remaining value-level `DataType` annotations and `isa(x, DataType)` checks to also accept `Type{T}` values. These changes keep the on-disk format identical to older Julia versions.

This change was made with the assistance of generative AI (Claude Code).

### DIFF
--- a/src/data/writing_datatypes.jl
+++ b/src/data/writing_datatypes.jl
@@ -75,7 +75,7 @@ function fieldodr(::Type{T}, initialized::Bool) where T
         if !hasfielddata(T)
             # A ghost type, so no need to store at all
             return nothing
-        elseif isa(T, DataType) && sizeof(T) ≤ MAX_INLINE_SIZE
+        elseif (isa(T, DataType) || Base.isType(T)) && sizeof(T) ≤ MAX_INLINE_SIZE
             if isbitstype(T)
                 return odr(T)
             elseif !ismutabletype(T)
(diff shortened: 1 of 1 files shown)


JonasIsensee:
Thanks for providing the update!

Codecov complains that the `isdefined(Core, :TypeEq) == true` branch is not hit as part of the tests.
Is this change not yet merged into julia nightly ? 

```
+ @static if isdefined(Core, :TypeEq)
+     @nospecializeinfer typeof_serializable(@nospecialize(x)) =
+         isa(x, Core.TypeEq) ? DataType : typeof(x)
+ else
+     @nospecializeinfer typeof_serializable(@nospecialize(x)) = typeof(x)
+ end
```

Keno:
> Codecov complains that the `isdefined(Core, :TypeEq) == true` branch is not hit as part of the tests.
> Is this change not yet merged into julia nightly ?

Merged after the PR, but before your comment ;)

Keno:
That said, the nightly build with it probably isn't done yet.

JonasIsensee:
Let's wait until then, retrigger CI, and merge when tests pass with the relevant code change covered.

src/data/writing_datatypes.jl
```diff
@@ -75,7 +75,7 @@ function fieldodr(::Type{T}, initialized::Bool) where T
         if !hasfielddata(T)
             # A ghost type, so no need to store at all
             return nothing
-        elseif isa(T, DataType) && sizeof(T) ≤ MAX_INLINE_SIZE
+        elseif (isa(T, DataType) || Base.isType(T)) && sizeof(T) ≤ MAX_INLINE_SIZE
```
nhz2, line 78: I don't think a change is needed here because this is inside an `if isconcretetype(T)` block.

nhz2:
```suggestion
        elseif isa(T, DataType) && sizeof(T) ≤ MAX_INLINE_SIZE
```

nhz2:
The tests are still missing coverage for these changes.

Examples of things still broken:

Unions are not saved correctly.

```julia-repl
julia> jldsave("union_new.jld2"; x = Union{Nothing, Type{Int}})

julia> load("union_new.jld2")
  "x" => Nothing

```

`Type{Int}` values cannot be read in older julia versions.

In Julia nightly
```julia-repl
julia> jldsave("datatypes_new.jld2"; x = Type{Int})

julia> load("datatypes_new.jld2")
  "x" => Type{Int64}

```

Loading this file in Julia 1.12.6
```julia-repl
julia> load("datatypes_new.jld2")
┌ Warning: custom serialization of Core.TypeEq encountered, but the type does not exist in the workspace; the data will be read unconverted
└ @ JLD2 ~/.julia/packages/JLD2/qcxKY/src/data/reconstructing_datatypes.jl:112
  "x" => UnknownType{Symbol("Core.TypeEq"), Tuple{Int64}}

```

There is a 2x file size regression storing arrays of `Type{Int}`, not sure this is practical problem, but it does show there was some change to the representation.

1.12.6:
```julia-repl
julia> jldsave("typearray.jld2"; x = fill(Type{Int}, 100000))

julia> filesize("typearray.jld2")
8801445

```

nightly:
```julia-repl
julia> jldsave("typearray.jld2"; x = fill(Type{Int}, 100000))

julia> filesize("typearray.jld2")
16379568

```

PR #726, opened 2026-06-30
Use workspaces and test filter downgrades

nhz2:
This PR cleanups the tests and uses workspaces for tests docs and the filter packages.

Older JLD2 versions and older filter packages need to be cross-compatible. A new test script `filter-downgrade-tests.jl` tests older versions of JLD2 and older versions of filter packages with the dev versions.


nhz2:
I'm going to merge this. It is correctly showing the current issues with the released lz4 filter constructor.

PR #727, opened 2026-07-02
Filters: Ignore extra client data values and document constructor

nhz2:
The LZ4 plugin in HDF5 is getting a new client value. Ref: https://github.com/HDFGroup/hdf5_plugins/pull/267

JLD2 should ignore client values it doen't know by default to stay compatible.

This PR adds a `from_client_values` constructor with this behavior. Filter plugins can specialize this function if the default behavior doesn't work well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.6.5
+ - Update OrderedCollections compat
+
 ## 0.6.4
  - Support for directly reading from and writing to Vector{UInt8}
  - Fix for working with opened IOStream objects
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JLD2"
 uuid = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
-version = "0.6.5"
+version = "0.6.6-dev"
 
 [deps]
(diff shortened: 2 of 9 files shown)


JonasIsensee:
Could we do this generically ?

```julia
# old 
F(fil.client_data...)

# idea
F(fil.client_data[min(end, fieldcount(F))]...)
```

As I see it, that would not require changes in the filter libraries for now.

nhz2:
Yes, that would work. It couples a few different things (which is maybe a good idea).
1. There is a direct correspondence between filter fields and client data values.
2. Extra client data values are always ignored, filter plugins have no way to see these.
3. The user facing default filter constructor is the same as a filter made from an empty client data values.

nhz2:
Okay the suggestion looks good.

nhz2:
Since this doesn't require changes to filter packages any more I'll go ahead and release JLD2Lz4 v0.1.2

nhz2:
The mapping between fields and cd_values may be tricky for ZFP https://h5z-zfp.readthedocs.io/en/latest/cd_vals.html

nhz2:
I can add back `from_client_values` but have it fall back to `F(fil.client_data[min(end, fieldcount(F))]...)`. That way filters that want to treat client values as some opaque array can specialize `from_client_values` and the regular filters still don't need any changes.

PR #729, opened 2026-07-06
Release JLD2Lz4 0.1.2

nhz2:
The v0.1.1 is missing the filter keyword arg constructor.

### DIFF
--- a/filter-downgrade-tests.jl
+++ b/filter-downgrade-tests.jl
@@ -1,9 +1,9 @@
 ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0
 
-pkgs_under_test = [
-    (; name="JLD2",      old_version=v"0.6.4", localpath=".")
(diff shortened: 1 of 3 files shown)


filter-downgrade-tests.jl
```diff
@@ -20,21 +20,33 @@ function run_out_tree_tests(;devs=nothing, adds=nothing)
     run(`$(Base.julia_cmd()) --project=$(test_dir) $(joinpath(test_dir, "runtests.jl"))`)
 end
 
-# First copy out the different packages under test to break the workspace
-pkg_temp_dirs = map(pkgs_under_test) do (;name, old_version, localpath)
-    cp(joinpath(@__DIR__, localpath), joinpath(mktempdir(), name))
+# Run the old JLD2's own test suite with the current filter packages swapped
+# into its source tree.
```
nhz2, line 24: Before I was testing the old version of JLD2 with the new test suite which wasn't quite right. With this the old test suite and old version of JLD2 are tested with the new filter packages.

PR #739, opened 2026-08-23
Fix ZstdFilter ignoring the requested compression level

giordano:
apply_filter! passed the level to ZstdEncodeOptions as the keyword `level`, but ChunkCodecLibZstd names it `compressionLevel` and its constructor silently discards unknown keywords, so every ZstdFilter compressed at the default level regardless of the requested one.

The options construction is factored out into a new `Filters.encode_options(filter)` function, so the parameter plumbing is directly testable: the regression test checks that the options constructed for ZstdFilter and Deflate carry the requested level, and that non-default levels round-trip through the file API.


src/Filters.jl
```diff
@@ -41,6 +41,14 @@ Depending on the filter type, this may include something like a compression leve
 """
 client_values(::Filter) = ()
 
+"""
+    encode_options(filter::Filter)
+
+Construct the `ChunkCodecCore.EncodeOptions` used to encode data with `filter`,
+carrying over the filter's parameters (such as the compression level).
+"""
+function encode_options end
```
giordano: This function is a bit silly, but it's only for making testing a little bit easier.

test/compression.jl
```diff
@@ -103,6 +103,30 @@ end
...
 end
 
+@testset "Compression level is applied" begin
+    # Regression test: ZstdEncodeOptions names its keyword `compressionLevel`
+    # and silently discards unknown keywords, so passing the level as `level`
+    # made every ZstdFilter compress at the default level regardless of the
+    # requested one. Check that the encode options constructed for a filter
+    # actually carry the requested level.
+    F = JLD2.Filters
+    for level in (1, 3, 22)
```
nhz2:
```suggestion
    for level in (-100, 1, 3, 22)
```
to also test negative levels

src/Filters.jl
nhz2: I think this function should be move to line 260 so it is more clear it is a helper function and not part of the `Filter` interface.

